PR #2, opened 2017-09-20
MAKE-266: avoid running jobs more than once with remote-backed queues

tychoish:


queue/remote_base.go
```diff
@@ -90,25 +91,22 @@ func (q *remoteBase) Complete(ctx context.Context, j amboy.Job) {
 	timer := time.NewTimer(0)
 	defer timer.Stop()
 
+	id := j.ID()
 	q.mutex.Lock()
-	delete(q.blocked, j.ID())
+	delete(q.blocked, id)
+	delete(q.dispatched, id)
```
bsamek: I don't understand how this interacts with `canDispatch`, which saves a `struct{}{}` at `q.dispatch[id]`. Why does this delete the id?

queue/remote_base.go
```diff
@@ -90,25 +91,22 @@ func (q *remoteBase) Complete(ctx context.Context, j amboy.Job) {
...
 	q.mutex.Unlock()
 
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		case <-timer.C:
-			err := q.driver.Lock(j)
```
bsamek: Why no longer take the lock?

PR #4, opened 2017-09-27
MAKE-270: update gonum/graph

tychoish:


tychoish:
This is really just a step on top of #3, put here for easy access and reference, it's just updating a vendored library.

PR #6, opened 2017-10-16
MAKE-280: ensure jobs are uniqe and that adding duplicate job ids is an error

tychoish:


queue/driver/driver_test.go
```diff
@@ -93,6 +93,21 @@ func (s *DriverSuite) TestInitialValues() {
 	s.Equal(0, stats.Total)
 }
 
+func (s *DriverSuite) TestPutJobDoesNotAllowDuplicateIds() {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	s.NoError(s.driver.Open(ctx))
+
+	j := job.NewShellJob("echo foo", "")
```
bsamek: nit: This tests adding identical jobs, not adding jobs with duplicate IDs, which is a superset. Construct two jobs that differ _except_ in their IDs to test duplicate IDs adds fail.

queue/driver/internal.go
```diff
@@ -70,6 +70,23 @@ func (d *Internal) Get(name string) (amboy.Job, error) {
...
 }
 
+// Put saves a new job to the queue, returning if it already exists.
+func (d *Internal) Put(j amboy.Job) error {
+	d.jobs.Lock()
+	defer d.jobs.Unlock()
+	name := j.ID()
+
+	_, a := d.jobs.m[name]
+	_, b := d.jobs.dispatched[name]
```
bsamek: Also check pending? I don't fully understand the difference between `d.jobs.m`, `d.jobs.dispatched`, and `d.jobs.pending`.

queue/driver/priority.go
```diff
@@ -70,11 +70,16 @@ func (p *Priority) Get(name string) (amboy.Job, error) {
...
 // an error.
 func (p *Priority) Save(j amboy.Job) error {
-	p.storage.Push(j)
+	p.storage.Save(j)
 
 	return nil
 }
 
+// Put saves a new job returning an error if that job already exists.
+func (p *Priority) Put(j amboy.Job) error {
```
bsamek: How do `Put` and `Save` differ?

PR #7, opened 2017-10-17
MAKE-276: implement panic recovery for job execution

tychoish:


pool/group.go
```diff
@@ -171,26 +172,46 @@ func (r *Group) Start(ctx context.Context) error {
...
+		// if we hit a panic we want to add an error to the job;
+		err = recovery.HandlePanicWithError(recover(), nil, "worker process encountered error")
+		if err != nil {
+			if unit.j != nil {
+				unit.j.AddError(err)
+				unit.q.Complete(ctx, unit.j)
+			}
+			// start a replacement worker.
+			wg.Add(1)
+			go groupWorker(ctx, wg, name, work)
```
bsamek: Couldn't this continue infinitely, each groupWorker panicking and spawning a new new one?

pool/group.go
```diff
@@ -171,26 +172,46 @@ func (r *Group) Start(ctx context.Context) error {
...
 	return nil
 }
 
+func groupWorker(ctx context.Context, wg *sync.WaitGroup, name string, work <-chan workUnit) {
+	var (
+		err  error
+		unit workUnit
+	)
+
+	defer wg.Done()
```
bsamek: Where is the Wait?

tychoish: in the close function for the pool

pool/group.go
tychoish: we run this only when there's a panic (e.g. 190)

PR #13, opened 2017-11-10
MAKE-300: local ordered-observing queue

tychoish:


queue/adaptive_order.go
```diff
@@ -0,0 +1,223 @@
...
+
+				if len(items.ready) > 0 {
+					id, items.ready = items.ready[0], items.ready[1:]
+					ret <- items.jobs[id]
+					return
+				}
+
+				items.refilter(ctx)
+
+				if len(items.ready) > 0 {
```
bsamek: Why do this twice?

PR #15, opened 2017-11-15
MAKE-310: account for size of pool when calculating time

tychoish:


pool/rate_limiting_average.go
```diff
@@ -72,18 +72,17 @@ func (p *ewmaRateLimiting) getNextTime(dur time.Duration) time.Duration {
...
+	runtimeOfTargetNumber := averageRuntime * time.Duration(p.target) / time.Duration(p.size)
+
+	// if jobs will finish in the period, we should do the following:
+	// 1. figure out how much extra time there is in the period (p.period - runtimeOfTargetNumber)
+	// 2. divide this by the number of targets to find wait time per target
+	// 3. multiply by the number of workers, since each worker should wait (waitTimePerTarget * numWorkers)
 	if runtimeOfTargetNumber < p.period {
-		workerRestingTime := p.period - runtimeOfTargetNumber
-		return workerRestingTime / time.Duration(p.target)
+		return (p.period - runtimeOfTargetNumber) / time.Duration(p.target) * time.Duration(p.size)
```
tychoish: - feels like we shouldn't be able to get a duration from this division, but maybe I'm wrong and its rounding internally
- we divide by size on the top and multiply by it on the bottom of the final expression, which seems weird. 
- the fact that make sure that we're larger than the period on line 84, seems wrong, given that we ought to account for size here too.

bsamek: > feels like we shouldn't be able to get a duration from this division, but maybe I'm wrong and its rounding internally

There is no rounding. Everything is an integer in this expression. It's a duration (`p.period - runtimeOfTargetNumber`) divided by a constant and then multiplied by a constant. The reason for casting them to durations is to make Go's type system happy, since it won't let you multiply `time.Duration`s by integers (though `time.Duration`s are `int64`s internally).

- we divide by size on the top and multiply by it on the bottom of the final expression, which seems weird. 

We divide in order to find out the total remaining time to run. We then multiply the diff by size. This allows us to exit early if the expected runtime is less than 0. We could fold 78 into 85 and then return 0 if the result is negative.

- the fact that make sure that we're larger than the period on line 84, seems wrong, given that we ought to account for size here too.

See my reply to your previous bullet point.

pool/rate_limiting_test.go
```diff
@@ -154,3 +156,23 @@ func TestEWMARateLimitingWorkerHandlesPanicingJobs(t *testing.T) {
...
+			period: time.Minute,
+			target: 60,
+			size:   workers,
+			queue:  nil,
+			ewma:   ewma.NewMovingAverage(),
+		}
+		for i := 0; i < 100; i++ {
+			next := ema.getNextTime(time.Millisecond)
+
+			if !assert.True(next > 750*time.Millisecond) || !assert.True(next < time.Second) {
```
bsamek: The sleep should increase as the number of workers increases. I believe my solution was correct.

PR #22, opened 2018-02-01
MAKE-340: add support for delayed start (waituntil) jobs

tychoish:


queue/adaptive_order_storage.go
```diff
@@ -57,6 +64,13 @@ func (items *adaptiveOrderItems) updateWaiting(ctx context.Context) {
 			continue
 		}
 		status := job.Status()
+		ti := job.TimeInfo()
+
+		if ti.WaitUntil.After(now) {
+			new = append(new, id)
```
bsamek: Is the inverse possible, that is, some item with an expired WaitUntil had its WaitUntil updated to be later? Or are items' WaitUntils fixed once they enter the queue? If the former, then you should also check items not in items.waiting? If the latter, then this is fine.

queue/smoke_test.go
```diff
@@ -219,6 +219,70 @@ func runOrderedSmokeTest(ctx context.Context, q amboy.Queue, size int, startBefo
...
+				assert.Zero(ti.WaitUntil)
+				assert.NoError(q.Put(j),
+					fmt.Sprintf("(a) with %d workers", num))
+				_, ok := q.Get(j.ID())
+				assert.True(ok)
+
+				cmd = fmt.Sprintf("echo %s.%d.b", name, num)
+				j = job.NewShellJob(cmd, "")
+				j.UpdateTimeInfo(amboy.JobTimeInfo{
+					WaitUntil: time.Now().Add(time.Hour),
```
bsamek: Is there a test for jobs with expired WaitUntils?

queue/adaptive_order_storage.go
tychoish: in general, jobs aren't mutable after the enter the queue, though for the non-remote queues it's just a pointer and it can be modified through the normal means. 

The adaptive order queue, reexamines the "blocked/waiting" queue, intermittently in the background to avoid stranding jobs that could be worked on in the queue. I think its fine in this case.

queue/smoke_test.go
tychoish: No, but a zero time is in the past, relative to now, so it should be the same?

PR #31, opened 2018-02-21
MAKE-348: improve marshal roundtrip test coverage

tychoish:


registry/raw.go
```diff
@@ -34,8 +34,28 @@ func (j *rawJob) GetBSON() (interface{}, error) { // Get ~= Marshal
...
+	job := factory()
+
+	err = um(job)
+	if err != nil {
+		return err
+	}
+
+	j.job = job
 
+	j.Body, err = amboy.ConvertTo(amboy.YAML, job)
```
bsamek: Shouldn't this be in the MarshalYAML() function?

registry/raw.go
```diff
@@ -91,7 +126,26 @@ func (d *rawDependency) GetBSON() (interface{}, error) { // Get ~= Marshal
...
+	}
+
+	dep := factory()
+	if err = um(dep); err != nil {
+		return errors.WithStack(err)
+	}
+
+	d.dep = dep
+
+	d.Body, err = amboy.ConvertTo(amboy.YAML, dep)
```
bsamek: same question

registry/raw.go
tychoish: Doing this makes the tests pass, though your right it shouldn't be needed.

Without this I think you couldn't roundtrip and then move to a different serialization method, because the yaml library is ~bonkers

PR #32, opened 2018-02-22
MAKE-352: optionally disable new queue query pending higher level fix

tychoish:


queue/driver_mongodb.go
```diff
@@ -400,7 +408,7 @@ func (d *mongoDB) Next(ctx context.Context) amboy.Job {
 			if err := query.One(j); err != nil {
 				misses++
 
-				if err == mgo.ErrNotFound || misses < 30 {
+				if err == mgo.ErrNotFound {
```
bsamek: log an error is misses exceeds some number?

PR #34, opened 2018-03-15
MAKE-359: add new bson library support

tychoish:


registry/formats.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "problem serializing data")
+	}
+
+	return output, nil
+
+}
+
+// ConvertFrom takes a Format type, a byte sequence, and an interface
+// and attempts to serialize that data into the interface object as
```
bsamek: deserialize

PR #36, opened 2018-03-26
MAKE-368: (CI) use appropriate task statuses

tychoish:


evergreen.yaml
```diff
@@ -205,18 +214,20 @@ tasks:
...
         - "gopath/src/github.com/mongodb/amboy/build/output.*"
   - command: shell.exec
+    type: setup
     params:
       script: |
         rm -rf amboy
         rm -rf ~/.aws
   - command: s3.put
-    type: system
+    type: setup
```
richardsamuels: at least in Evergreen, we kept s3.put as system.

PR #38, opened 2018-03-28
MAKE-372: task completion time should be included in task duration

tychoish:


pool/local.go
```diff
@@ -127,15 +128,17 @@ func worker(ctx context.Context, jobs <-chan amboy.Job, q amboy.Queue, wg *sync.
 				Start: time.Now(),
 			}
 
+			job.UpdateTimeInfo(ti)
```
bsamek: after 134

tychoish: calling it twice intentionally.

bsamek: lgtm when this is fixed

PR #47, opened 2018-04-04
MAKE-385: change lock updating strategy in driver

tychoish:


queue/driver_locker.go
```diff
@@ -90,10 +90,11 @@ func (l *lockManager) lockPinger(ctx context.Context) {
 					continue
 				}
 
-				// Update the important metadata
 				stat := j.Status()
-				stat.ModificationCount++
-				stat.ModificationTime = time.Now()
+				if !stat.InProgress || stat.Owner != l.name {
```
bsamek: Doest this mean, don't run this job?

PR #48, opened 2018-04-04
MAKE-391: avoid dispatching stale jobs

tychoish:


pool/rate_limiting.go
```diff
@@ -170,5 +170,13 @@ func (p *simpleRateLimited) Close() {
 	if p.canceler != nil {
 		p.canceler()
 	}
+
+	// because of the timer+2 contexts in the worker
+	// implementation, we can end up returning earlier and because
+	// pools are restartable, end up calling wait more than once,
+	// which doesn't affect behavior but does cause this to panic in
+	// tests
+	defer func() { recover() }()
```
bsamek: If this has access to a logger, will you log anyway, just in case we need to debug some problem that turns out to be a panic?

PR #51, opened 2018-06-14
MAKE-416: fix test race conditions

tychoish:


pool/rate_limiting.go
```diff
@@ -178,5 +199,7 @@ func (p *simpleRateLimited) Close() {
 	// tests
 	defer func() { recover() }()
 
+	// to let workers cancel
+	time.Sleep(10 * time.Millisecond)
```
bsamek: Short waits in close functions seems wrong.

PR #52, opened 2018-07-02
MAKE-374: queue reporting package

tychoish:


queue/driver_mongodb.go
```diff
@@ -395,30 +395,42 @@ func (d *mongoDB) Next(ctx context.Context) amboy.Job {
...
-			{
-				"status.completed": false,
-				"status.in_prog":   true,
-				"status.mod_ts":    bson.M{"$lte": time.Now().Add(-lockTimeout)},
-			},
-		},
-	}
-
-	if d.respectWaitUntil {
+	if d.useNewQuery {
```
bsamek: What's this change?

tychoish: I think this was wrong in earlier versions, and we cleaned up the default.

PR #53, opened 2018-07-05
MAKE-425: abort should call complete in the queue

tychoish:


develar:
Thanks a lot, works :)

PR #54, opened 2018-07-06
MAKE-426 log errors at error level

john-m-liu:


tychoish:
This is fine, but there is a generic grip.Log(<level>, <message>) method meant for this kind of thing.

PR #57, opened 2018-07-10
MAKE-419 MAKE-427: add abortable features to rate limiting pool and cleanup cached distributed locks

tychoish:


pool/abortable.go
```diff
@@ -123,21 +124,29 @@ func (p *abortablePool) worker(ctx context.Context, jobs <-chan amboy.Job) {
 				job.AddError(err)
 				p.queue.Complete(ctx, job)
 			}
+
 			// start a replacement worker.
 			go p.worker(ctx, jobs)
 		}
+
+		if cancel != nil {
```
john-m-liu: How can cancel not be nil here?

pool/helpers.go
```diff
@@ -75,24 +81,34 @@ func worker(ctx context.Context, jobs <-chan amboy.Job, q amboy.Queue, wg *sync.
 			// start a replacement worker.
 			go worker(ctx, jobs, q, wg)
 		}
+
+		if cancel != nil {
```
john-m-liu: How can cancel not be nil here?

pool/rate_limiting.go
```diff
@@ -119,6 +120,9 @@ func (p *simpleRateLimited) worker(ctx context.Context, jobs <-chan amboy.Job) {
 			// start a replacement worker.
 			go p.worker(ctx, jobs)
 		}
+		if cancel != nil {
```
john-m-liu: How can cancel not be nil here?

pool/rate_limiting_average.go
```diff
@@ -153,6 +156,9 @@ func (p *ewmaRateLimiting) worker(ctx context.Context, jobs <-chan amboy.Job) {
 			// start a replacement worker.
 			go p.worker(ctx, jobs)
 		}
+		if cancel != nil {
```
john-m-liu: How can cancel not be nil here?

pool/rate_limiting_average.go
```diff
@@ -181,6 +191,22 @@ func (p *ewmaRateLimiting) runJob(ctx context.Context, j amboy.Job) time.Duratio
 	}
 	j.UpdateTimeInfo(ti)
 
+	var cancel context.CancelFunc
+	ctx, cancel = context.WithCancel(ctx)
+	func() {
```
john-m-liu: Did you intend to run this in a separate thread?

tychoish: no it's just here for a scope so that we can defer unlocking the mutex.

pool/abortable.go
tychoish: if the jobs channel only gets nil jobs, and then the context is canceled, this will run before cancel is set and we panic.

pool/rate_limiting_average.go
john-m-liu: Could you make it a named function instead of an anonymous one?

PR #58, opened 2018-07-11
MAKE-425: additional lock cancelation tests

tychoish:


queue/driver_locker.go
```diff
@@ -98,7 +98,29 @@ func (l *lockManager) lockPinger(ctx context.Context) {
 				}
 
 				stat := j.Status()
-				if !stat.InProgress || stat.Owner != l.name {
+				if !stat.InProgress {
```
bsamek: is `!state.InProgress` really the same thing as a complete job? I don't know what the other state are.

queue/driver_locker.go
```diff
@@ -108,42 +130,51 @@ func (l *lockManager) lockPinger(ctx context.Context) {
...
+	l.ops <- func(pings lockPings) {
 		pings[name] = lockPingOp{
 			ts:  time.Now().Add(l.timeout),
 			ctx: ctx,
 		}
 		close(wait)
 	}
 
-	select {
-	case <-ctx.Done():
```
bsamek: Why no context check?

queue/driver_locker_test.go
```diff
@@ -25,16 +24,13 @@ func TestLockManagerSuite(t *testing.T) {
...
 	s.driver = NewPriorityDriver().(*priorityDriver)
-	ctx, s.suiteCancel = context.WithCancel(context.Background())
-	s.Require().NoError(s.driver.Open(ctx))
 }
 
 func (s *LockManagerSuite) SetupTest() {
 	s.ctx, s.testCancel = context.WithCancel(context.Background())
 	s.lm = newLockManager("test", s.driver)
-	s.lm.timeout = 100 * time.Millisecond
+	s.lm.timeout = 300 * time.Millisecond
```
bsamek: lol

queue/remote_base.go
```diff
@@ -121,17 +123,26 @@ func (q *remoteBase) Complete(ctx context.Context, j amboy.Job) {
...
 					return
+				} else if count > 10 {
+					grip.Error(message.WrapError(err, message.Fields{
+						"job_id":      id,
+						"job_type":    j.Type().Name,
+						"driver_type": fmt.Sprintf("%T", q.driver),
+						"driver_id":   q.driver.ID(),
+						"retry_count": count,
+						"message":     " after 10 retries, aborting marking job complete",
+					}))
```
bsamek: You need a return here

queue/driver_locker.go
tychoish: we have a job, we lock it (e.g. set it in progress) and then we put it in the ping list, which is what this list is. 

We should keep pinging it until it's no longer locked, (e.g. not in progress) regardless of if it was marked complete or not, so I'm pretty sure this is correct.

queue/driver_locker.go
tychoish: the context is here to cancel the lock ping once it starts, not to cancel the act of adding a lock to the pinger. I think its less complicated to do this and was causing problems, I'd like to hold off on this change until later. just in terms of making this easier to update in evergreen.

queue/driver_locker_test.go
tychoish: eh.

queue/remote_base.go
tychoish: done

PR #63, opened 2018-10-26
Make 422

ablack12:


queue/sqs.go
```diff
@@ -0,0 +1,260 @@
...
+
+// Used to mark a Job complete and remove it from the pending
+// work of the queue.
+func (q *sqsFIFOQueue) Complete(ctx context.Context, job amboy.Job) {
+	name := job.ID()
+	q.mutex.Lock()
+	defer q.mutex.Unlock()
+	if ctx.Err() != nil {
+		grip.Noticef("did not complete %s job, because operation "+
+			"was canceled.", name)
```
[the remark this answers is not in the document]
ablack12: Do you mean Logf? What are the different priority levels?

queue/sqs.go
```diff
@@ -0,0 +1,260 @@
...
+		return
+	}
+	q.tasks.completed[name] = true
+	if q.tasks.all[name] != nil {
```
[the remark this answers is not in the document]
ablack12: Not actually sure what you mean by this. Do you mean save q.tasks.all[name] to a variable? Or something else?

queue/sqs.go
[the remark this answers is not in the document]
tychoish: We should use a message.Fields (which is just a map[string]interface{} for reporting so that the messages are easier to parse in splunk/etc.

queue/sqs.go
```diff
@@ -70,26 +69,31 @@ func (q *sqsFIFOQueue) Put(j amboy.Job) error {
...
+		return errors.Wrap(err, "Error converting job in Put")
+	}
 	job, err := json.Marshal(jobItem)
-	if err != nil { // is this the right way to handle this?
-		return err
+	if err != nil {
+		return errors.Wrap(err, "Error marshalling job to JSON in Put")
 	}
-	groupID := RandomString(16)
-	dedupID := j.ID()
```
ablack12: Amazon errors if the deduplication ID has spaces (accepts alphanumeric values and punctuation). Is this okay? I could have something to change spaces in the ID to underscores or something but that seems hacky. I don't know what job IDs usually are though.

tychoish: they should be unique strings. you could just trim all whitespace characters.

queue/sqs.go
```diff
@@ -17,9 +17,12 @@ import (
 	"github.com/mongodb/amboy/registry"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
+	"github.com/mongodb/grip/recovery"
 	"github.com/pkg/errors"
 )
 
+const region string = "us-west-2"
```
tychoish: why is this hardcoded? shouldn't this be us-east?

queue/sqs.go
```diff
@@ -189,7 +184,7 @@ func (q *sqsFIFOQueue) Results(ctx context.Context) <-chan amboy.Job {
 			if _, ok := q.tasks.completed[name]; ok {
 				select {
 				case <-ctx.Done():
-					return
+					recovery.LogStackTraceAndContinue("results context is done")
```
tychoish: this isn't right. we shouldn't expect a panic here.

ablack12:
45 commits later, this can probably be looked at again.

PR #67, opened 2019-01-04
MAKE-595: run job handler

tychoish:


meta.go
```diff
@@ -77,3 +80,36 @@ func Report(ctx context.Context, q Queue, limit int) QueueReport {
...
+	job.Run(ctx)
+	ti.End = time.Now()
+	msg := message.Fields{
+		"job":           job.ID(),
+		"job_type":      job.Type().Name,
+		"duration_secs": ti.Duration().Seconds(),
+	}
+	err := errors.WithStack(job.Error())
+	if err != nil {
+		grip.Error(message.WrapError(err, msg))
```
bsamek: You could ErrrorWhen instead of if/else.

PR #70, opened 2019-02-12
MAKE-641 MAKE-655: cleanup logging and pool operations

tychoish:


pool/abortable.go
```diff
@@ -171,7 +174,25 @@ func (p *abortablePool) runJob(ctx context.Context, job amboy.Job) {
 		delete(p.jobs, job.ID())
 	}()
 
-	handleJob(ctx, job, p.queue)
+	start := time.Now()
+
+	executeJob(ctx, job, p.queue, start)
+
+	ti := job.TimeInfo()
```
bsamek: Nit: You only use this in one place, so no need to assign to a variable here.

pool/abortable.go
```diff
@@ -171,7 +174,25 @@ func (p *abortablePool) runJob(ctx context.Context, job amboy.Job) {
...
+	r := message.Fields{
+		"job":           job.ID(),
+		"job_type":      job.Type().Name,
+		"duration_secs": ti.Duration().Seconds(),
+		"queue_type":    fmt.Sprintf("%T", p.queue),
+		"pool_type":     "abortable",
+	}
+	if err := job.Error(); err != nil {
+		r["error"] = err.Error()
+		grip.Error(r)
```
bsamek: You can use `<level>When` and get rid of the if statement if you wanted.

tychoish:
merged

PR #72, opened 2019-03-15
MAKE-686 Create local QueueGroup type

bsamek:


queue/group.go
```diff
@@ -0,0 +1,132 @@
...
+	"context"
+	"sync"
+	"time"
+
+	"github.com/mongodb/amboy"
+	"github.com/pkg/errors"
+)
+
+// NewLocalQueueGroup constructs a new local queue group. If ttl is 0, the collection will not be
+// TTLed except when the client explicitly calls Prune.
```
tychoish: collection isn't what you really mean, I think?

queue/group.go
```diff
@@ -0,0 +1,132 @@
...
+func NewLocalQueueGroup(ctx context.Context, constructor QueueConstructor, ttl time.Duration) (amboy.QueueGroup, error) {
+	if constructor == nil {
+		return nil, errors.New("must pass a constructor")
+	}
+	if ttl < 0 {
```
tychoish: I think we should probably validate at 100ms or even 1s

queue/group.go
```diff
@@ -0,0 +1,132 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	g := &localQueueGroup{
+		canceler:    cancel,
+		queues:      map[string]amboy.Queue{},
+		constructor: constructor,
+		ttlMap:      map[string]time.Time{},
+		ttl:         ttl,
+	}
+	if ttl > 0 {
+		go func(ctx context.Context) {
```
tychoish: recovery handler. 

also I don't think you need to pass context here because it's in the outer scope

queue/group.go
```diff
@@ -0,0 +1,132 @@
...
+					g.Prune()
+				}
+			}
+		}(ctx)
+	}
+	return g, nil
+}
+
+// QueueConstructor is a function passed by the client which makes a new queue.
+type QueueConstructor func(ctx context.Context) (amboy.Queue, error)
```
tychoish: should we move this file to a different package? the name is weird, as most queue constructors have a different signature

queue/group.go
```diff
@@ -0,0 +1,132 @@
...
+	g.ttlMap[id] = time.Now()
+	return queue, nil
+}
+
+// Put a queue at the given index.
+func (g *localQueueGroup) Put(id string, queue amboy.Queue) error {
+	g.mu.RLock()
+	if _, ok := g.queues[id]; ok {
+		g.mu.RUnlock()
+		return errors.WithStack(errors.New("a queue already exists at this index"))
```
tychoish: you don't have to withstack new errors from the error package.

queue/group.go
```diff
@@ -0,0 +1,132 @@
...
+	return nil
+}
+
+// Close the queues.
+func (g *localQueueGroup) Close(ctx context.Context) {
+	g.canceler()
+	wg := &sync.WaitGroup{}
+	for _, queue := range g.queues {
+		wg.Add(1)
+		go func(queue amboy.Queue) {
```
tychoish: panic handler, and defer t he done

queue/group.go
```diff
@@ -0,0 +1,132 @@
...
+// Prune old queues.
+func (g *localQueueGroup) Prune() error {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	for queueID, t := range g.ttlMap {
+		if time.Since(t) > g.ttl {
+			if _, ok := g.queues[queueID]; ok {
+				delete(g.queues, queueID)
+			}
+			delete(g.ttlMap, queueID)
```
tychoish: delete noops if the key doesn't exist, so you can probably remove the ifstatment

queue/group.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}
+	return nil
+}
+
+// Close the queues.
+func (g *localQueueGroup) Close(ctx context.Context) {
+	g.canceler()
```
tychoish: you could probably test that prune calls cancelr by defining a func() that you set to g.canceler

queue/group.go
bsamek: Done

queue/group.go
bsamek: Done

queue/group.go
bsamek: Not anymore, since I gather the queues to close

queue/group.go
bsamek: I tried this, but I couldn't get it to close over a variable in the outer scope for the test to work.

queue/group.go
bsamek: Done.

interface.go
```diff
@@ -157,6 +157,23 @@ type Queue interface {
...
+	Get(context.Context, string) (Queue, error)
+
+	// Put a queue at the given index.
+	Put(string, Queue) error
+
+	// Prune old queues.
+	Prune() error
+
+	// Close the queues.
+	Close(context.Context)
```
tychoish: should this error?

queue/group.go
```diff
@@ -0,0 +1,150 @@
...
+	wg := &sync.WaitGroup{}
+	for _, queue := range g.queues {
+		wg.Add(1)
+		go func(queue amboy.Queue) {
+			defer wg.Done()
+			defer recovery.LogStackTraceAndContinue("panic in local queue group closer")
```
john-m-liu: Better to reverse the order of this I think?

PR #77, opened 2019-04-08
EVG-5887 Fix up remote queue group

bsamek:


queue/group_remote.go
```diff
@@ -16,6 +16,8 @@ import (
 	"gopkg.in/mgo.v2/bson"
 )
 
+const jobsSuffix = ".jobs"
```
tychoish: can we use this constant in the driver implementations that get collection names? and maybe have utility functions to add/strip?

bsamek: done, though in a somewhat different way

queue/util.go
```diff
@@ -11,3 +12,11 @@ func randomString(x int) string {
 	_, _ = rand.Read(b)
 	return hex.EncodeToString(b)
 }
+
+func AddJobsSuffix(s string) string {
```
tychoish: make these private.

PR #79, opened 2019-04-12
EVG-5887 Fix up remote queue group

bsamek:


queue/group_remote.go
```diff
@@ -108,7 +116,7 @@ func NewRemoteQueueGroup(ctx context.Context, opts RemoteQueueGroupOptions) (amb
 				case <-ctx.Done():
 					return
 				case <-ticker.C:
-					_ = g.Prune(ctx)
+					grip.Error(errors.Wrap(g.Prune(pruneCtx), "problem pruning remote queue group database"))
```
tychoish: why not collect this and propogate out? catchers are thread safe. 

also message.WrapError() is probably also good/better?

queue/group_remote.go
```diff
@@ -250,21 +261,24 @@ func (g *remoteQueueGroup) Prune(ctx context.Context) error {
...
-	for i, coll := range colls {
+	collsToCheck := []string{}
+	for _, coll := range colls {
 		// This is an optimization. If we've added to the queue recently enough, there's no
 		// need to query its contents, since it cannot be old enough to prune.
 		if t, ok := g.ttlMap[g.idFromCollection(coll)]; ok && time.Since(t) < g.ttl {
-			g.remove(colls, i)
+			// pass
+		} else {
+			collsToCheck = append(collsToCheck, coll)
```
tychoish: invert the logic to avoid the else?

queue/group_remote.go
```diff
@@ -250,21 +261,24 @@ func (g *remoteQueueGroup) Prune(ctx context.Context) error {
...
 	catcher := grip.NewBasicCatcher()
 	wg := &sync.WaitGroup{}
-	collsDeleteChan := make(chan string, len(colls))
+	collsDeleteChan := make(chan string, len(collsToCheck))
 	collsDropChan := make(chan string)
 
 	go func() {
 		defer recovery.LogStackTraceAndContinue("panic in pruning collections")
-		for _, coll := range colls {
+		for _, coll := range collsToCheck {
```
tychoish: given that this collsDropChan is buffered, do we need to do this in a thread?

queue/group_remote.go
```diff
@@ -286,7 +300,7 @@ func (g *remoteQueueGroup) Prune(ctx context.Context) error {
 					catcher.Add(err)
 					return
 				}
-				if err := one.Decode(struct{}{}); err != nil {
+				if err := one.Decode(&struct{}{}); err != nil {
```
tychoish: given that we don't care about the result, shouldn't we just be doing counts here?

queue/group_remote.go
bsamek: This goroutine keeps running after the constructor exits.

queue/group_remote.go
bsamek: done

queue/group_remote.go
bsamek: done

queue/group_remote.go
bsamek: done

PR #81, opened 2019-04-17
EVG-6001: legacy driver implementation to queue group

tychoish:


tychoish:
one test times out, @bsamek (the prune-with-ttl) thoughts may be useful.

bsamek:
@tychoish It looks like the queues are not running the jobs. However, `q.Start` is called, so I don't know what the problem is.

PR #83, opened 2019-04-19
MAKE-746: queue groups in one collection

tychoish:


queue/driver_group_mongo.go
```diff
@@ -0,0 +1,520 @@
...
+		respectWaitUntil: opts.CheckWaitUntil,
+		instanceID:       fmt.Sprintf("%s.%s.%s.%s", name, group, host, uuid.NewV4()),
+	}
+}
+
+// OpenNewMongoGroupDriver constructs and opens a new MongoDB driver instance
+// using the specified session. It is equivalent to calling
+// NewMongoGroupDriver() and calling driver.Open().
+func OpenNewMongoGroupDriver(ctx context.Context, name string, opts MongoDBOptions, group string, groupTTL time.Duration, client *mongo.Client) (Driver, error) {
+	d := NewMongoGroupDriver(name, opts, group, groupTTL).(*mongoGroupDriver)
```
bsamek: nit: Check this type assertion and return an error if it fails.

queue/group_remote_mongo_single.go
```diff
@@ -0,0 +1,226 @@
...
+				case <-ticker.C:
+					grip.Error(message.WrapError(g.Prune(pruneCtx), "problem pruning remote queue group database"))
+				}
+			}
+		}()
+	}
+
+	if opts.BackgroundCreateFrequency > 0 {
+		go func() {
+			defer recovery.LogStackTraceAndContinue("panic in remote queue group ticker")
```
bsamek: These defers and the one in the above goroutine are inconsistently ordered. I think the panic handler should be the first on the stack.

PR #90, opened 2019-04-23
MAKE-687: use common queue caching mechanism

tychoish:


queue/group_utils.go
```diff
@@ -0,0 +1,232 @@
...
+
+type cacheItem struct {
+	q    amboy.Queue
+	ttl  time.Duration
+	ts   time.Time
+	name string
+}
+
+func (c *cacheImpl) Len() int {
+	c.mu.Lock()
```
bsamek: I think this can be an RLock()

queue/group_utils.go
```diff
@@ -0,0 +1,232 @@
...
+	defer c.mu.Unlock()
+
+	return len(c.q)
+}
+
+func (c *cacheImpl) Names() []string {
+	c.mu.Lock()
```
bsamek: RLock()

queue/group_utils.go
```diff
@@ -0,0 +1,232 @@
...
+
+					if item.ttl > 0 && item.ttl > time.Since(item.ts) {
+						continue
+					}
+
+					if item.q.Stats().IsComplete() {
+						item.q.Runner().Close(ctx)
+
+						wait := make(chan struct{})
+						func() {
```
bsamek: Did you mean to put this in a goroutine?

tychoish: nope, want to be able to lock and use a defer but have it happen synchronously.

tychoish: jk, you're right.

PR #98, opened 2019-05-11
MAKE-687: refactor smoke tests

tychoish:


queue/remote_ordered.go
```diff
@@ -57,21 +58,23 @@ func (q *remoteSimpleOrdered) Next(ctx context.Context) amboy.Job {
 				Start: time.Now(),
 			}
 			job.UpdateTimeInfo(ti)
-			err := q.driver.Lock(ctx, job)
```
bsamek: Why remove this?

queue/queue_test.go
```diff
@@ -0,0 +1,995 @@
+package queue
```
bsamek: I can't actually tell what changed because of the file rename. Maybe let's talk in person for a minute.

queue/remote_ordered.go
tychoish: it just moves to line 75, after we check it...

this is also only in the ordered queue.

PR #99, opened 2019-05-13
EVG-6150 Don't mark complete if context is canceled

bsamek:


bsamek:
also see https://github.com/evergreen-ci/evergreen/pull/2298

PR #102, opened 2019-05-14
MAKE-744: support for queue groups

tychoish:


reporting/mongodb.go
```diff
@@ -40,6 +42,27 @@ func NewDBQueueState(ctx context.Context, name string, opts queue.MongoDBOptions
 	return db, nil
 }
 
+// NewDBQueueGroupState produces a queue Reporter for a (remote)
+// queue group queue. This implementation does not interact with a queue
+// directly, and reports by interacting with the database directly.
+func NewDBQueueGroupState(ctx context.Context, name string, opts queue.MongoDBOptions, group string) (Reporter, error) {
```
bsamek: This seems reasonable, but I think we might want another interface, one that finds all queue group queues and groups them.

tychoish: I avoided doing that. I think if you do the standard one but against a group queue, then it mostly works, at least treating all queues as one (which works for a lot of the operations here) I think you also want a different interface for aggregated stats across queues, and maybe with different questions?

PR #104, opened 2019-05-30
MAKE-603: amboy noop pool

tychoish:


pool/noop.go
```diff
@@ -0,0 +1,50 @@
...
+	if p.queue == nil {
+		return errors.New("cannot start pool without set queue")
+	}
+
+	p.isStarted = true
+	return nil
+}
+
+func (p *noopPool) SetQueue(q amboy.Queue) error {
+	if q == nil {
```
bsamek: This is mutexed in other implementations.

tychoish: I think you only need the mutex in other implementations because the queue is used from the worker threads, but here it never would be, so it's not an issue and won't trip the race detector (which is why we added the mutex, I think the race is really theoretical.) 

I'm inclined to avoid the mutex unless the race detector fails.

PR #105, opened 2019-06-05
MAKE-603: interval queue operation with option to skip

tychoish:


periodic.go
```diff
@@ -19,10 +19,15 @@ type QueueOperation func(context.Context, Queue) error
...
+// The theshold, if ResepectThreshold is set, causes the periodic
+// scheduler to noop if there are more than that many pending jobs.
 type QueueOperationConfig struct {
-	ContinueOnError bool `bson:"continue_on_error" json:"continue_on_error" yaml:"continue_on_error"`
-	LogErrors       bool `bson:"log_errors" json:"log_errors" yaml:"log_errors"`
-	DebugLogging    bool `bson:"debug_logging" json:"debug_logging" yaml:"debug_logging"`
+	ContinueOnError  bool `bson:"continue_on_error" json:"continue_on_error" yaml:"continue_on_error"`
+	LogErrors        bool `bson:"log_errors" json:"log_errors" yaml:"log_errors"`
+	DebugLogging     bool `bson:"debug_logging" json:"debug_logging" yaml:"debug_logging"`
+	RespectThreshold bool `bson:"respect_threshold" json:"respect_threshold" yaml:"respect_threshold"`
```
bsamek: This interface is a little awkward. Could you use 0 to indicate that `scheduleOp` should ignore the threshold. That's awkward too, but maybe less so.

PR #106, opened 2019-06-13
MAKE-794 Add index to support remote group getQueues

bsamek:


tychoish:
should we do this to all the driver implementations?

bsamek:
> should we do this to all the driver implementations?

done

PR #109, opened 2019-07-08
MAKE-829: enable not having TTL indexes

julianedwards:


tychoish:
there should probably be some documentation of this in the options struct. also it should be more clear to users that they have to express TTL in terms of seconds.

queue/driver_group_mongo.go
```diff
@@ -188,7 +183,10 @@ func (d *mongoGroupDriver) setupDB(ctx context.Context) error {
 				},
 			},
 		},
-		mongo.IndexModel{
+	}
+	if d.opts.TTL > 0 {
```
bsamek: @tychoish Should we have a way of opting the queue out of the TTL index? Since we use duplicate key exceptions to prevent re-executing queue group jobs, a TTL will cause sufficiently old tasks to fail if restarted.

julianedwards: Talked offline: these options are created per queue and therefore can be customized for each queue. If a queue does not want a TTL index it just needs to leave the TTL option at 0 and no index will be created.

tychoish: The group actually shares a single collection and set of indexes. There's no viable way to have a group with different ttl values.

PR #110, opened 2019-07-09
MAKE-836: add dispatch by checks to most queues

tychoish:


interface.go
```diff
@@ -89,20 +90,46 @@ type JobStatusInfo struct {
...
+	DispatchBy time.Time     `bson:"dispatch_by" json:"dispatch_by,omitempty" yaml:"dispatch_by,omitempty"`
+	MaxTime    time.Duration `bson:"max_time" json:"max_time,omitempty" yaml:"max_time,omitempty"`
 }
 
 // Duration is a convenience function to return a duration for a job.
 func (j JobTimeInfo) Duration() time.Duration { return j.End.Sub(j.Start) }
 
+// IsStale determines if the job is too old to be dispatched, and if
+// so, queues may remove or drop the job entirely.
+func (j JobTimeInfo) IsStale() bool { return j.DispatchBy.After(time.Now()) }
```
bsamek: Shouldn't this be `Before`? If I want something dispatched by tomorrow, it's not stale if tomorrow is _later_ than today.

interface.go
```diff
@@ -89,20 +90,46 @@ type JobStatusInfo struct {
...
+
+// IsDispatchable determines if the job should be delayed based on the
+// value of WaitUntil.
+func (j JobTimeInfo) IsDispatchable() bool { return j.WaitUntil.After(time.Now()) }
```
bsamek: Shouldn't this also be `Before`?

interface.go
```diff
@@ -89,20 +90,46 @@ type JobStatusInfo struct {
...
+
+// Validate ensures that the structure has reasonable values set.
+func (j JobTimeInfo) Validate() error {
+	catcher := grip.NewBasicCatcher()
+
+	catcher.NewWhen(!j.DispatchBy.IsZero() && j.WaitUntil.After(j.DispatchBy), "invalid for wait_until to be after dispatch_by")
+	catcher.NewWhen(j.Created.IsZero(), "must specify non-zero created timestamp")
+	catcher.NewWhen(j.MaxTime < 0, "must specify positive")
```
bsamek: "must specific 0 or positive MaxTime"

queue/adaptive_order_storage.go
```diff
@@ -85,6 +85,11 @@ func (items *adaptiveOrderItems) updateWaiting(ctx context.Context) {
 			continue
 		}
 
+		if ti.DispatchBy.After(now) {
```
bsamek: Use `IsStale`

queue/adaptive_order_storage.go
```diff
@@ -120,6 +125,11 @@ func (items *adaptiveOrderItems) updateStalled(ctx context.Context) {
 			continue
 		}
 
+		if job.TimeInfo().DispatchBy.After(time.Now()) {
```
bsamek: `Use IsStale`

queue/priority.go
```diff
@@ -62,14 +67,26 @@ func (q *priorityLocalQueue) Get(ctx context.Context, name string) (amboy.Job, b
...
+				grip.Notice(message.Fields{
+					"state":    "stale",
+					"job":      job.ID(),
+					"job_type": job.Type().Name,
+				})
+				continue
+			}
+
+			q.counters.Lock()
+			defer q.counters.Unlock()
```
bsamek: This is in a `for` loop, so don't use a `defer`.

bsamek:
Also, can you write tests?

queue/adaptive_order_storage.go
tychoish: I want to use the same "now" value, hence this one being weird.

interface.go
tychoish: no, because you want it to be in the past in order to dispatch.

interface.go
tychoish: I think we have to negate in order to handle 0 values for dispatch by correct, but yeah

bsamek: I don't think you changed this.

interface.go
bsamek: I want to wait until some time. Now that time is yesterday, so I've waited until it. Yesterday is _not_ after now.

queue/adaptive_order_storage.go
bsamek: You could make a parameterized helper for `IsStale`.

interface.go
```diff
@@ -89,20 +90,46 @@ type JobStatusInfo struct {
...
+	DispatchBy time.Time     `bson:"dispatch_by" json:"dispatch_by,omitempty" yaml:"dispatch_by,omitempty"`
+	MaxTime    time.Duration `bson:"max_time" json:"max_time,omitempty" yaml:"max_time,omitempty"`
 }
 
 // Duration is a convenience function to return a duration for a job.
 func (j JobTimeInfo) Duration() time.Duration { return j.End.Sub(j.Start) }
 
+// IsStale determines if the job is too old to be dispatched, and if
+// so, queues may remove or drop the job entirely.
+func (j JobTimeInfo) IsStale() bool { return !j.DispatchBy.After(time.Now()) }
```
bsamek: This doesn't work with zero time. The beginning of time is not after now and so the zero value is stale, which will cause unset values to be removed or dropped. I'm not sure there's anything you can do except deliberately check zero time.

queue/driver_group_mgo.go
```diff
@@ -396,16 +396,16 @@ func (d *mgoGroupDriver) Next(ctx context.Context) amboy.Job {
...
 
+	timeLimits := bson.M{}
+	now := time.Now()
 	if d.opts.CheckWaitUntil {
-		qd = bson.M{
-			"$and": []bson.M{
-				qd,
-				{"$or": []bson.M{
-					{"time_info.wait_until": bson.M{"$lte": time.Now()}},
-					{"time_info.wait_until": bson.M{"$exists": false}}},
```
bsamek: Did you deliberately remove this part of the $or?

queue/driver_group_mgo.go
```diff
@@ -396,16 +396,16 @@ func (d *mgoGroupDriver) Next(ctx context.Context) amboy.Job {
...
-				},
-			},
-		}
+		timeLimits["time_info.wait_until"] = bson.M{"$lte": now}
+	}
+	if d.opts.CheckDispatchBy {
+		timeLimits["time_info.dispatch_by"] = bson.M{"$gt": now}
```
bsamek: Does this need an index?

queue/driver_group_mgo.go
tychoish: yes, I believe that it is no longer required.

queue/driver_group_mgo.go
tychoish: perhaps, for now it's going to be off by default, and we can revisit as we turn on. I imagine it'll be ok to no use an index, but we can see.

tychoish:
The deleted tests (which I will change to skip are just slow cases that I want to disable in more places. 

Will write tests.

interface.go
```diff
@@ -89,20 +90,54 @@ type JobStatusInfo struct {
...
+// so, queues may remove or drop the job entirely.
+func (j JobTimeInfo) IsStale() bool {
+	if j.DispatchBy.IsZero() {
+		return false
+	}
+
+	return j.DispatchBy.Before(time.Now())
+}
+
+// IsDispatchable determines if the job should be delayed based on the
```
bsamek: "should be delayed" -> "can be dispatched"

queue/driver_group_mgo.go
```diff
@@ -396,16 +396,16 @@ func (d *mgoGroupDriver) Next(ctx context.Context) amboy.Job {
...
-		qd = bson.M{
-			"$and": []bson.M{
-				qd,
-				{"$or": []bson.M{
-					{"time_info.wait_until": bson.M{"$lte": time.Now()}},
-					{"time_info.wait_until": bson.M{"$exists": false}}},
-				},
-			},
-		}
+		timeLimits["time_info.wait_until"] = bson.M{"$lte": now}
```
bsamek: Does this need to check that it's not 0? (similar question a couple more times below too)

tychoish: if it's zero, then it'll be less than now, so I think it's fine.

PR #111, opened 2019-07-09
MAKE-837: avoid abandoning stale queues

tychoish:


queue/group_remote_mgo_single.go
```diff
@@ -114,6 +114,10 @@ func (g *remoteMgoQueueGroupSingle) getQueues(ctx context.Context) ([]string, er
 						"status.completed": true,
 						"status.mod_ts":    bson.M{"$gte": time.Now().Add(-g.opts.TTL)},
 					},
+					{
+						"status.in_prog": true,
```
bsamek: This isn't supported by an index.

tychoish: Given the `$or` query, this might be a touch complicated, and it might be ok to do some document scanning given that it's probably running a few times an hour, and the number of in_prog:true jobs is <1000.

bsamek: There isn't even an index on "status.in_prog" though. We should add that.

tychoish: this raises an interesting question: we're already querying for status.completed:false which should cover all of the in_progress queues, which means this shouldn't ever happen.. I think this fix/bug is probably not right.

PR #112, opened 2019-07-24
MAKE-864: improve unlock behavior

tychoish:


queue/driver_locker.go
```diff
@@ -73,8 +73,7 @@ func (l *lockManager) lockPinger(ctx context.Context) {
 		case op := <-l.ops:
 			op(activeLocks)
 		case <-timer.C:
-			startAt := time.Now()
-			nextLoopAt := time.Now().Add(l.timeout / 2)
+			nextLoopAt := time.Now().Add(l.timeout)
```
bsamek: Why remove the `/2`?

tychoish: wanted to make this less aggressive.

PR #113, opened 2019-08-07
MAKE-883: save error count

tychoish:


bsamek:
You should either delete the `ErrorCount()` function or have it return the field.

PR #114, opened 2019-08-22
MAKE-901: update locking interface

tychoish:


pool/helpers.go
```diff
@@ -50,6 +56,40 @@ func runJob(ctx context.Context, job amboy.Job, q amboy.Queue, startAt time.Time
...
+	}
+	if err := q.Save(ctx, job); err != nil {
+		job.AddError(errors.Wrap(err, "problem saving job state"))
+		return false
+	}
+
+	pingerCtx, stopPing := context.WithCancel(ctx)
+	go func() {
+		iters := 0
+		timer := time.NewTimer(amboy.LockTimeout / 2)
```
bsamek: This would be slightly less code with a ticker, and potentially less error-prone with a ticker. (You don't have to call `Reset`.)

interface.go
```diff
@@ -146,9 +153,13 @@ func (j JobTimeInfo) Validate() error {
 // organization properties.
 type Queue interface {
 	// Used to add a job to the queue. Should only error if the
-	// Queue cannot accept jobs.
+	// Queue cannot accept jobs or if the job already exists in a
```
bsamek: Delete "or"

pool/helpers.go
```diff
@@ -50,6 +56,40 @@ func runJob(ctx context.Context, job amboy.Job, q amboy.Queue, startAt time.Time
...
+	if err := job.Lock(q.ID()); err != nil {
+		job.AddError(errors.Wrap(err, "problem locking job"))
+		return false
+	}
+	if err := q.Save(ctx, job); err != nil {
+		job.AddError(errors.Wrap(err, "problem saving job state"))
+		return false
+	}
+
+	pingerCtx, stopPing := context.WithCancel(ctx)
```
bsamek: `defer stopPing()` instead of calling at the end of the function

pool/mock_queue_test.go
```diff
@@ -43,12 +46,32 @@ func (q *QueueTester) Put(ctx context.Context, j amboy.Job) error {
...
 		return nil
 	}
 }
 
+func (q *QueueTester) Save(ctx context.Context, j amboy.Job) error {
+	q.mutex.Lock()
+	defer q.mutex.Unlock()
+
+	id := j.ID()
+	if _, ok := q.storage[id]; !ok {
```
bsamek: I read this as, "If it's not already in storage, don't store it." That seems wrong.

queue/sqs.go
```diff
@@ -107,8 +117,27 @@ func (q *sqsFIFOQueue) Put(ctx context.Context, j amboy.Job) error {
...
+
+	q.mutex.Lock()
+	defer q.mutex.Unlock()
+
+	if !q.Started() {
+		return errors.Errorf("cannot save job %s; queue not started", name)
+	}
+
+	if _, ok := q.tasks.all[name]; !ok {
+		return errors.Errorf("cannot add %s because duplicate job does not exist", name)
```
bsamek: I don't understand what this means.

pool/helpers.go
tychoish: my instinct is that you want to be sure that this is not running before you call complete.

pool/mock_queue_test.go
tychoish: save shouldn't save jobs that aren't tracked (use put for that) and put shouldn't save jobs that are tracked (use put.) I think this is correct.

pool/helpers.go
tychoish: but I think you can call cancel more than once so it's all good!

pool/mock_queue_test.go
tychoish: the production implementations return an error here, but in practice for in memory queues you don't need to call save because the job is always the same object.

tychoish:
I responded to the feedback, and also layered in some changes that should get the full build green: 
- reducing the size of the test load for windows/osx, we should get full coverage, but will run fewer tests and more of the smaller tests. 
- turned off a repeated test-execution builder on the race detector.
- tightened down timeouts to be more realistic. (from 1h to 30 or 15m) 
- 1 or 2 small test fixes that had been occluded by the larger timeout issue.

I also, on a lark, listed multiple windows and osx distros in my patch, to pilot the new distro alias functionality and it worked flawlessly, so I left that in too. 

Test times are down a couple of minutes on the Linux builder as well, which I think bodes well. I'm confident in the existing multiple-execution tests that the change won't cause a regression for short running tests. The current structure of the code doesn't make it easy to test "jobs that run for longer than the timeout," but there's a ticket for that, and this code does include some changes to the job-end-logging that will make it easier to introspect cases where we get locking related issues.

PR #131, opened 2019-11-15
MAKE-923: queue group periodic operation

tychoish:


periodic_group.go
```diff
@@ -0,0 +1,112 @@
...
+
+		waitUntilInterval(ctx, startAt, interval)
+
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
+
+		if ctx.Err() != nil {
+			return
+		}
+		count := 1
```
Kimchelly: nit: can we move this count to the for loop below this where it's actually used?

periodic_group.go
```diff
@@ -0,0 +1,112 @@
...
+
+// IntervalGroupQueueOperation schedules jobs on a queue group with
+// similar semantics as IntervalQueueOperation.
+//
+// Operations will continue to run as long as the context is not
+// canceled. If you do not pass any GroupQueueOperation items to this
+// function, it panics.
+func IntervalGroupQueueOperation(ctx context.Context, qg QueueGroup, interval time.Duration, startAt time.Time, conf QueueOperationConfig, ops ...GroupQueueOperation) {
+	if len(ops) == 0 {
+		panic("queue group operation must contain operations")
```
Kimchelly: Is it okay that we panic instead of just return an error here?

periodic_group.go
```diff
@@ -0,0 +1,112 @@
...
+		q, err := group.Get(ctx, op.Queue)
+		if conf.ContinueOnError {
+			grip.WarningWhen(conf.LogErrors, err)
+		} else {
+			grip.CriticalWhen(conf.LogErrors, err)
+			return errors.Wrapf(err, "problem getting queue '%s' from group", op.Queue)
+		}
+
+		if err = scheduleOp(ctx, q, op.Operation, conf); err != nil {
+			return errors.Wrapf(err, "problem scheduling job on group queue '%s'", op.Queue)
```
Kimchelly: Since the goroutine in IntervalGroupQueueOperation returns without logging anything, should we log in this error case?

periodic_group.go
tychoish: these functions don't return errors, we could noop, but panicing seems save.

periodic_group.go
tychoish: if we need to log, it's logged in schedule op, so I took out the wrapping.

PR #132, opened 2020-01-08
MAKE-1086: add scopes

tychoish:


bsamek:
@tychoish Will you fix compile first? I started reviewing but then realized the things I was noticing will be caught by the compiler.

interface.go
```diff
@@ -79,6 +79,11 @@ type Job interface {
 	// of queue.ID()
 	Lock(string) error
 	Unlock(string)
+
+	// Scope provides the ability to provide more configurable
+	// exclusion a job can provide.
+	Scope() []string
```
bsamek: `Scopes`, not `Scope`, to match the implementation.

job/base.go
```diff
@@ -254,3 +255,21 @@ func (b *Base) UpdateTimeInfo(i amboy.JobTimeInfo) {
...
+
+	b.scopes = scopes
+}
+
+// Scopes returns the required scopes for the job.
+func (b *Base) Scopes() []string {
+	b.mutex.RLock()
+	defer b.mutex.RUnlock()
+
+	return b.timeInfo
```
bsamek: `return b.scopes`

tychoish:
compile is fixed but lint is broken beytond the PR

queue/adaptive_order.go
```diff
@@ -263,15 +266,31 @@ func (q *adaptiveLocalOrdering) Next(ctx context.Context) amboy.Job {
 
 				if len(items.ready) > 0 {
 					id, items.ready = items.ready[0], items.ready[1:]
-					ret <- items.jobs[id]
+					j := items.jobs[id]
+					if err := q.scopes.Acquire(id, j.Scopes()); err != nil {
+						items.ready = append(items.ready)
+						misses++
+						timer.Reset(time.Duration(misses * rand.Int63n(int64(time.Millisecond))))
```
bsamek: The timer should be reset to 0 if err is nil.

queue/adaptive_order.go
```diff
@@ -263,15 +266,31 @@ func (q *adaptiveLocalOrdering) Next(ctx context.Context) amboy.Job {
...
 				items.refilter(ctx)
 
 				if len(items.ready) > 0 {
 					id, items.ready = items.ready[0], items.ready[1:]
-					ret <- items.jobs[id]
+					j := items.jobs[id]
+					if err := q.scopes.Acquire(id, j.Scopes()); err != nil {
+						items.ready = append(items.ready)
+						misses++
+						timer.Reset(time.Duration(misses * rand.Int63n(int64(time.Millisecond))))
```
bsamek: Same comment about the timer.

queue/priority.go
```diff
@@ -102,6 +104,12 @@ func (q *priorityLocalQueue) Next(ctx context.Context) amboy.Job {
 				}()
 				continue
 			}
+			if err := q.scopes.Acquire(job.ID(), job.Scopes()); err != nil {
+				go func() {
+					defer recovery.LogStackTraceAndContinue("re-queue waiting job", job.ID())
+					q.channel <- job
```
bsamek: Why send a job in the channel if acquiring he lock fails?

queue/sqs.go
```diff
@@ -174,6 +176,10 @@ func (q *sqsFIFOQueue) Next(ctx context.Context) amboy.Job {
 	if err != nil {
 		return nil
 	}
+	if err := q.scopes.Acquire(job.ID(), job.Scopes()); err != nil {
+		q.Put(ctx, job)
```
bsamek: Why call q.Put when acquisition fails?

interface.go
```diff
@@ -79,6 +79,11 @@ type Job interface {
 	// of queue.ID()
 	Lock(string) error
 	Unlock(string)
+
```
bsamek: Write tests

queue/adaptive_order.go
tychoish: I think not actually... the property of missing a lock, and then having a little bit of jitter (not much really) is probably good. the max sleep is 10ms

queue/priority.go
tychoish: because otherwise it falls on the floor and is never acted upon.

queue/sqs.go
tychoish: because otherwise failing to get the lock (ususally because someone else has it) will cause us to never do this job.

interface.go
tychoish: the current test suite goes a long way to ensure that this patch won't regress in cases where locks don't have scope (and already ensure that the primary locking mechanism (i.e. that jobs only ever run once.) is sufficient, so even though the tests haven't changed, I think passing tests provide good signal here. 

There's a trivial case that I plan to write once I get a green patch that if I add scopes to a bunch of sleeping jobs and too many threads, that it'll not dispatch half of them, but those tests are always kind of terrible.

job/base.go
```diff
@@ -254,3 +255,29 @@ func (b *Base) UpdateTimeInfo(i amboy.JobTimeInfo) {
...
+
+	b.RequiredScopes = scopes
+}
+
+// Scopes returns the required scopes for the job.
+func (b *Base) Scopes() []string {
+	b.mutex.RLock()
+	defer b.mutex.RUnlock()
+
+	if len(b.RequiredScopes) == 0 {
```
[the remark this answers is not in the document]
tychoish: because we want nil so you can get the proper omitempty behavior.

queue/driver_mongo.go
```diff
@@ -454,38 +468,62 @@ func isMongoDupKey(err error) bool {
 }
 
 func (d *mongoDriver) Save(ctx context.Context, j amboy.Job) error {
-	name := j.ID()
+	job, err := d.prepareInterchange(j)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	job.Scopes = j.Scopes()
```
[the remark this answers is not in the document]
tychoish: because you want to be able to have different values for scope when you complete (so you effectively release the lock there.)

queue/dispatcher.go
```diff
@@ -0,0 +1,140 @@
...
+		return
+	}
+	delete(d.cache, job.ID())
+
+	ti := job.TimeInfo()
+	ti.End = time.Now()
+	job.UpdateTimeInfo(ti)
+
+	info.jobCancel()
+	info.stopPing()
```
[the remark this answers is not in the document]
tychoish: passed to different resources. I think I'd rather keep things the way it has been for now.

PR #133, opened 2020-01-29
MAKE-1133: Revendor grip (with updated Sender interface)

julianedwards:


logger/sender_multi.go
```diff
@@ -59,9 +62,28 @@ func NewQueueMultiSender(ctx context.Context, workers, capacity int, senders ...
 }
 
 func (s *multiQueueSender) Send(m message.Composer) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
tychoish: I think this can be an RLock

logger/sender_multi.go
```diff
@@ -2,14 +2,17 @@ package logger
...
 
 	"github.com/mongodb/amboy"
 	"github.com/mongodb/amboy/queue"
+	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
 	"github.com/mongodb/grip/send"
 )
 
 type multiQueueSender struct {
+	mu       sync.Mutex
```
tychoish: make rwlock

logger/sender_multi.go
julianedwards: how so? we are writing to the queue here

PR #152, opened 2020-04-15
MAKE-1259: dispatching fixes and logging

tychoish:


queue/driver_mongo.go
```diff
@@ -775,18 +775,20 @@ RETRY:
...
-						"job_id":    job.ID(),
-						"job_type":  job.Type,
-						"scopes":    job.Scopes,
-						"stat":      job.Status,
-						"is_group":  d.opts.UseGroups,
-						"group":     d.opts.GroupName,
-						"dup_key":   isMongoDupKey(err),
-					}))
+					grip.DebugWhen(
+						!isDispatchable(job.Status()),
```
julianedwards, line 779: shouldn't we debug when `isDispatchable` rather than `!isDispatchable`?

PR #158, opened 2020-04-28
MAKE-902: make LockTimeout configurable

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-902

* Add lock timeout setting for mongo driver options.
* Add `Info()` method to `Queue` interface that supercedes `Started()` and returns driver-configurable lock timeout.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -77,8 +77,8 @@ type Job interface {
 	// not-in-progress, and should be a no-op if the job does not
 	// belong to the owner. In general the owner should be the value
 	// of queue.ID()
-	Lock(string) error
-	Unlock(string)
+	Lock(owner string, lockTimeout time.Duration) error
+	Unlock(owner string, lockTimeout time.Duration)
 
 	// Scope provides the ability to provide more configurable
 	// exclusion a job can provide.
@@ -185,7 +185,11 @@ type Queue interface {
 
 	// Makes it possible to detect if a Queue has started
 	// dispatching jobs to runners.
+	// kim: TODO: remove and replace with Info()
 	Started() bool
+
+	// Info returns information related to management of the Queue.
+	Info() QueueInfo
 
 	// Used to mark a Job complete and remove it from the pending
 	// work of the queue.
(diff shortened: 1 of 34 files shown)


interface.go
```diff
@@ -223,6 +222,11 @@ type Queue interface {
 	Start(context.Context) error
 }
 
+type QueueInfo struct {
```
tychoish, line 225: docstrings

queue/driver.go
```diff
@@ -25,6 +25,9 @@ type remoteQueueDriver interface {
 	JobStats(context.Context) <-chan amboy.JobStatusInfo
 	Complete(context.Context, amboy.Job) error
 
+	// kim: TODO: maybe remove
```
tychoish, line 28: the queue drivers are all internal, and at some point we'll restructure the mongod queues to avoid the remote queue+driver vestige... In the mean time, I think it's fine.

queue/driver.go
```diff
@@ -48,6 +51,17 @@ type MongoDBOptions struct {
 	// field. If set to zero, the TTL index will not be created and
 	// and documents may live forever in the database.
 	TTL time.Duration
+	// LockTimeout overrides the default job lock timeout if set.
+	LockTimeout time.Duration
+}
+
+// GetLockTimeout returns the lock timeout specified by the options, or
+// amboy.LockTimeout as the default.
+func (opts *MongoDBOptions) GetLockTimeout() time.Duration {
```
tychoish, line 60: I might omit this method and just put this value in the DefaultMongoDBOptions constructor, and potentially have a validation method on MongoDBOptions, but not actually force it to be correct here.

interface.go
Kimchelly: Done.

queue/driver.go
Kimchelly: Done.

queue/driver.go
```diff
@@ -78,5 +69,22 @@ func DefaultMongoDBOptions() MongoDBOptions {
 		SkipReportingIndexBuilds: false,
 		WaitInterval:             time.Second,
 		Format:                   amboy.BSON,
+		LockTimeout:              amboy.LockTimeout,
+	}
+}
+
+// ValidateAndDefault validates that the required options are given and sets
+// fields that are unspecified and have a default value.
+func (opts *MongoDBOptions) ValidateAndDefault() error {
```
tychoish, line 78: I'd just call it Validate()

PR #159, opened 2020-04-28
MAKE-1215: re-dispatch job on lock timeout

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-1215

I'm not entirely confident that this is why barque's scope jobs don't restart when the app server restarts but my current theory is that when the job gets stuck in progress, `(*mongoDriver).getNextQuery()` finds jobs that have exceeded the lock timeout but they get filtered by `isDispatchable()` in `(*dispatcherImpl).Dispatch()`. This just allows jobs that have exceeded the lock timeout. to be re-dispatched.

### DIFF
--- a/queue/remote_base.go
+++ b/queue/remote_base.go
@@ -341,15 +341,12 @@ func (q *remoteBase) canDispatch(j amboy.Job) bool {
 	return true
 }
 
-func isDispatchable(stat amboy.JobStatusInfo) bool {
-	if stat.ModificationTime.Before(time.Now().Add(-amboy.LockTimeout)) {
-		return true
-	}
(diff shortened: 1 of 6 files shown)


queue/remote_base.go
```diff
@@ -337,6 +337,9 @@ func (q *remoteBase) canDispatch(j amboy.Job) bool {
 }
 
 func isDispatchable(stat amboy.JobStatusInfo) bool {
+	if stat.ModificationTime.Before(time.Now().Add(-amboy.LockTimeout)) {
```
tychoish, line 340: should we add the queue info as an argument here so dynamic lock timeout works?

PR #161, opened 2020-05-13
MAKE-911: benchmark queue throughput

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-911

I didn't add a task for it since it would probably take a while, but I could add it if it's useful.

* Vendor poplar.
* Benchmark throughput of mock jobs (number of jobs the queue can process in a fixed amount of time)
    * Use mongo-backed queue.
    * Benchmarks jobs with scopes and no scopes.
    * I can add more benchmark cases (e.g. N different scope jobs in the queue concurrently) if that would be useful.


benchmarks/harness.go
```diff
@@ -0,0 +1,231 @@
...
+			return errors.Wrap(err, "making queue")
+		}
+		defer func() {
+			grip.Error(errors.Wrap(closeQueue(), "cleaning up queue"))
+		}()
+
+		errChan := make(chan error)
+		qctx, cancel := context.WithCancel(ctx)
+		defer cancel()
+		go func() {
```
tychoish, line 77: should we defer close the error channel here?

benchmarks/harness.go
```diff
@@ -0,0 +1,231 @@
...
+	return catcher.Resolve()
+}
+
+type closeFunc func() error
+
+type queueConstructor func(context.Context) (amboy.Queue, closeFunc, error)
+type jobConstructor func(id string) amboy.Job
+
+func basicThroughputBenchmark(makeQueue queueConstructor, makeJob jobConstructor, timeout time.Duration) poplar.Benchmark {
+	return func(ctx context.Context, r poplar.Recorder, count int) error {
```
tychoish, line 65: are we using the count somehow?

cmd/run-benchmarks/run-benchmarks.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/mongodb/amboy/benchmarks"
+)
+
+func main() {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	err := benchmarks.RunQueue(ctx)
+	if err != nil {
+		fmt.Println(err)
```
tychoish, line 16: exit one here, or just grip.EmergencyFatal() the error

Kimchelly:
Results are put in the output directory in `benchmarks.RunQueue()`. The summary file is `results.txt`.

tychoish:
Cool, can the results.txt get uploaded to evergreen?

Kimchelly:
I think results.txt doesn't actually provide anything useful (like number of jobs processed). It only gives basic summary things like number of iterations and runtime. The actual useful data would be in the ftdc files.

tychoish:
We should also figure out some way to put the FTDC into cedar, so there can be charts eventually. Having said that it can happen outside the scope of this ticket.

What's here is LGTM

tychoish:
We should also figure out some way to put the FTDC into cedar, so there can be charts eventually. Having said that it can happen outside the scope of this ticket.

What's here is LGTM

PR #164, opened 2020-05-18
MAKE-1328: amboy registry cleanup

tychoish:


registry/raw.go
```diff
@@ -1,222 +1,21 @@
...
-	job  interface{}
-}
+type rawJob []byte
 
-func (j *rawJob) SetBSON(r legacyBSON.Raw) error { j.Body = r.Data; return nil }
-func (j *rawJob) GetBSON() (interface{}, error) { // Get ~= Marshal
-	if j.job != nil {
-		return j.job, nil
-	}
+func (j rawJob) MarshalJSON() ([]byte, error)   { return j, nil }
```
julianedwards, line 7: This can be a pointer... the json library, unlike the bson library, works with nicely with converting values to pointers.

registry/raw.go
```diff
@@ -1,222 +1,21 @@
...
-	var err error
-	d.Body, err = convertTo(amboy.BSON2, d.dep)
-
-	if err != nil {
-		return nil, errors.WithStack(err)
-	}
-
-	return d.Body, nil
-}
+func (d rawDependency) MarshalJSON() ([]byte, error)   { return d, nil }
```
julianedwards, line 16: same as above

PR #165, opened 2020-05-26
MAKE-1186: Allow aborting amboy jobs by prefix

julianedwards:


management/queue.go
```diff
@@ -581,3 +582,50 @@ func (m *queueManager) CompleteJobs(ctx context.Context, f StatusFilter) error {
...
+		if stat.Completed {
+			continue
+		}
+		if !strings.HasPrefix(stat.ID, prefix) {
+			continue
+		}
+
+		switch f {
+		case Stale:
+			if stat.InProgress && time.Since(stat.ModificationTime) > m.queue.Info().LockTimeout {
```
bsamek, line 604: I read this to pick out jobs that _are_ stale, whereas I read the `InProgress` case to pick out jobs that are _not_ in progress.

rest/service_management.go
```diff
@@ -37,6 +37,7 @@ func (s *ManagementService) App() *gimlet.APIApp {
 	app.AddRoute("/jobs/mark_complete/{name}").Version(1).Post().Handler(s.MarkComplete)
 	app.AddRoute("/jobs/mark_complete_by_type/{type}/{filter}").Version(1).Post().Handler(s.MarkCompleteByType)
 	app.AddRoute("/jobs/mark_many_complete/{filter}").Version(1).Post().Handler(s.MarkManyComplete)
+	app.AddRoute("/jobs/mark_complete_by_prefix/{prefix}/{filter}").Version(1).Post().Handler(s.MarkCompleteByPrefix)
```
bsamek, line 40: Is there a case where we wouldn't want to pass `{filter}`? Or alternatively, what's the behavior if `{filter} == ""`. Should it be all?

rest/service_management.go
```diff
@@ -227,6 +228,26 @@ func (s *ManagementService) MarkManyComplete(rw http.ResponseWriter, r *http.Req
...
+// MarkCompleteByPrefix is an http.Handlerfunc marks all jobs with the
+// specified prefix and status complete.
+func (s *ManagementService) MarkCompleteByPrefix(rw http.ResponseWriter, r *http.Request) {
+	vars := gimlet.GetVars(r)
+	prefix := vars["prefix"]
+	filter := vars["filter"]
+
+	ctx := r.Context()
+	if err := s.manager.CompleteJobsByPrefix(ctx, management.StatusFilter(filter), prefix); err != nil {
+		gimlet.WriteResponse(rw, gimlet.MakeTextErrorResponder(errors.Wrapf(err,
```
bsamek, line 244: Does it really make sense to wrap here?  Presumably we know where this error is coming from, and the thing we actually care about is the stack that we already put in `err`. If we wrap here, there will be two stacks in the error message, and I don't think we'll care about the one from this function.

julianedwards: I think this is more of a Sam thing. I am not super opinionated on this, but we have been wrapping everything else.

management/queue.go
julianedwards: I am actually confused now, I copied and pasted this... but looking at all the management functions they each do different things with the `Stale` case.

rest/service_management.go
julianedwards: yeah then the user should pass in `All`

management/queue.go
```diff
@@ -601,7 +601,7 @@ func (m *queueManager) CompleteJobsByPrefix(ctx context.Context, f StatusFilter,
 
 		switch f {
 		case Stale:
-			if stat.InProgress && time.Since(stat.ModificationTime) > m.queue.Info().LockTimeout {
+			if !stat.InProgress || time.Since(stat.ModificationTime) > m.queue.Info().LockTimeout {
```
bsamek, line 604: forgot to flip the `>`

PR #171, opened 2021-01-12
EVG-13717: add option to apply scopes on enqueue

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13717

* Add job option to make a job acquire scopes when it's enqueued (i.e. when `Put()` is called) rather than when the job is dispatched.
* Implement the option for the Mongo driver, which backs the remote queue implementation.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -84,6 +84,11 @@ type Job interface {
 	// exclusion a job can provide.
 	Scopes() []string
 	SetScopes([]string)
+
+	// ApplyScopesOnEnqueue allows the scope exclusion functionality to be
+	// configured so that exclusion occurs during job dispatch or enqueue.
+	ApplyScopesOnEnqueue() bool
+	SetApplyScopesOnEnqueue(bool)
 }
 
 // JobType contains information about the type of a job, which queues
--- a/interface.go
+++ b/interface.go
@@ -85,10 +85,10 @@ type Job interface {
 	Scopes() []string
 	SetScopes([]string)
 
-	// ApplyScopesOnEnqueue allows the scope exclusion functionality to be
+	// ShouldApplyScopesOnEnqueue allows the scope exclusion functionality to be
 	// configured so that exclusion occurs during job dispatch or enqueue.
-	ApplyScopesOnEnqueue() bool
-	SetApplyScopesOnEnqueue(bool)
+	ShouldApplyScopesOnEnqueue() bool
+	SetShouldApplyScopesOnEnqueue(bool)
 }
 
 // JobType contains information about the type of a job, which queues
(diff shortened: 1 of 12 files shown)


interface.go
```diff
@@ -84,6 +84,11 @@ type Job interface {
 	// exclusion a job can provide.
 	Scopes() []string
 	SetScopes([]string)
+
+	// ApplyScopesOnEnqueue allows the scope exclusion functionality to be
+	// configured so that exclusion occurs during job dispatch or enqueue.
+	ApplyScopesOnEnqueue() bool
```
julianedwards, line 90: This is a bit confusing cause `apply` here can also sound like a verb. I almost want to change this to `IsApplyScopesOnEnqueueSet` or `IsApplyScopesOnEqueue` but maybe that is too long. What do you think?

queue/driver_test.go
```diff
@@ -102,6 +104,91 @@ func (s *DriverSuite) TestPutJobDoesNotAllowDuplicateIds() {
...
+	j = job.NewShellJob("echo bar", "")
+	j.SetScopes([]string{"scope"})
+	j.SetApplyScopesOnEnqueue(true)
+
+	s.Require().NoError(s.driver.Put(s.ctx, j))
+
+	s.Equal(2, s.driver.Stats(s.ctx).Total)
+}
+
+func (s *DriverSuite) TestPutAndSaveJobSucceedsIfScopeIsAppliedOnEnqueue() {
```
julianedwards, line 175: What is the difference between this test and `TestPutJobAllowsSameScopesInQueueIfInitialScopedJobDoesNotApplyScopeOnEnqueue`?

queue/remote_test.go
```diff
@@ -69,14 +61,15 @@ func TestRemoteUnorderedMongoSuite(t *testing.T) {
 // above Test function and replacing the driverConstructor function.
 
 func (s *RemoteUnorderedSuite) SetupSuite() {
-	s.require = s.Require()
+
```
julianedwards, line 64: Can we remove this function?

julianedwards (changes requested):

queue/driver_test.go
Kimchelly: Sorry, the other test was a copy-pasta mistake. I updated it to test the actual `Put()`/`Save()` sequence.

queue/remote_test.go
Kimchelly: Removed.

interface.go
julianedwards: I agree, `Should` makes the most sense.

Kimchelly: Done.

julianedwards (approved):

PR #172, opened 2021-01-14
EVG-13538: add option for retryable job

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13538

Implement retryable options with a flag indicating if the job can retry and how many times it's retried. They're no-ops until [EVG-13540](https://jira.mongodb.org/browse/EVG-13540) is implemented to have a retry handler, but they are serialized so they can be stored in MongoDB.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -144,8 +144,8 @@ type JobTimeInfo struct {
 // JobRetryInfo stores configuration and information for a job that can retry.
 // Support for retrying jobs is optional for Queue implementations.
 type JobRetryInfo struct {
-	Retryable  bool `bson:"retryable,omitempty" json:"retryable,omitempty" yaml:"retryable,omitempty"`
-	RetryCount int  `bson:"retry_count,omitempty" json:"retry_count,omitempty" yaml:"retry_count,omitempty"`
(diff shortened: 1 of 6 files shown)


interface.go
```diff
@@ -129,6 +136,13 @@ type JobTimeInfo struct {
 	MaxTime    time.Duration `bson:"max_time" json:"max_time,omitempty" yaml:"max_time,omitempty"`
 }
 
+// JobRetryInfo stores configuration and information for a job that can retry.
+// Support for retrying jobs is optional for Queue implementations.
+type JobRetryInfo struct {
```
julianedwards, line 141: Why do we need both `Retryable` and `RetryCount`? I think having a single configuration called `RetryCount` or `Retries` would be simpler.

julianedwards (changes requested):

interface.go
Kimchelly: `Retryable` tells you whether the job can retry at all. `RetryCount` is not meant to be modified by you - it tells you how many times it's tried in total so far.

julianedwards: Ohh I thought it was like the max retry attempts allowed. Can we set a max?

Kimchelly: [EVG-13544](https://jira.mongodb.org/browse/EVG-13544)

julianedwards (approved):

PR #173, opened 2021-01-15
EVG-13718: implement applying scope on enqueue for local scope manager implementations

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13718

* Make the local scope manager operations atomic. Previously, it was possible to partially acquire/release the scopes that you pass in.
* Implement applying scopes for implementations that already use a local scope manager.
* Make the queue implementations consistent with the remote queue/Mongo driver, which acquires the scope in `Save()`.
* Write common queue interface tests for all implementations that have scopes.
* Fix some test coverage issues where a bunch of implementations that support WaitUntil/DispatchBy wouldn't test them.

### DIFF
--- a/queue/scope.go
+++ b/queue/scope.go
@@ -6,7 +6,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-// ScopeManager provides a service to queue implementation to support
+// ScopeManager provides a service to queue implementations to support
 // additional locking semantics for queues that cannot push that into
 // their backing storage.
 type ScopeManager interface {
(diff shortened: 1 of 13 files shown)


queue/scope.go
```diff
@@ -48,6 +49,10 @@ func (s *scopeManagerImpl) Acquire(id string, scopes []string) error {
 		return errors.Errorf("could not acquire lock scope '%s' held by '%s' not '%s'", sc, holder, id)
 	}
 
+	for _, sc := range scopesToAcquire {
```
julianedwards, line 52: I am confused why you moved this down here instead of keeping it in the for loop above?

Kimchelly: If you don't do this, you can end up partially acquiring all of the scopes you asked for. If you ask for the "foo" and "bar" scopes but "bar" is already in use by a different job, the old code would successfully acquire "foo" and error trying to acquire "bar". This change is to make acquiring all of the scopes atomic (and i wrote a test for it) - either you get all the scopes you ask for, or none of them.

julianedwards (approved):

PR #174, opened 2021-01-25
EVG-13760: cancel the job's context after MaxTime elapses

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13760

MaxTime currently does nothing if you set it except set an error message after the job completes. It should now behave as expected and cancel the job's `Run()` context after MaxTime (which must be configured before the job has been enqueued).

### DIFF
--- a/queue/queue_test.go
+++ b/queue/queue_test.go
@@ -54,6 +54,7 @@ type QueueTestCase struct {
 	OrderedStartsBefore     bool
 	WaitUntilSupported      bool
 	DispatchBeforeSupported bool
+	MaxTimeSupported        bool
 	ScopesSupported         bool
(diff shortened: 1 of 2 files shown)


queue/queue_test.go
```diff
@@ -783,6 +797,40 @@ waitLoop:
 	assert.Equal(t, size.Size, stats.Completed)
 }
 
+/* kim: TODO: test
```
julianedwards, line 800: Should we resolve this?

Kimchelly: Oh yeah, I wrote the test, I just forgot to remove it.

julianedwards (approved):

PR #178, opened 2021-02-08
EVG-13652: release dispatched jobs that error before they execute

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13652

A job got dispatched and the lock ping thread started, but the job never ran. The only places I could see where a job could be dispatched without running were a few exceptional error cases in between the dispatcher starting the lock ping thread and the `job.Run()` call. I made it so that all the exit conditions in between those two events cause the dispatcher to release the job. Since I didn't see any of the exceptional error logs in Splunk, I'm not confident that one of these errors was the cause, but the logging should be better now so we can diagnose it better if it recurs.

* Release any job that successfully gets dispatched but errors before it can run the job.  
* Refactor the dispatch attempt loop to be more readable.
* Add more logging at all the error cases after dispatch occurs but before the job gets run.

### DIFF
--- a/queue/dispatcher.go
+++ b/queue/dispatcher.go
@@ -55,6 +55,12 @@ func (d *dispatcherImpl) Dispatch(ctx context.Context, job amboy.Job) error {
 	defer d.mutex.Unlock()
 
 	if _, ok := d.cache[job.ID()]; ok {
+		grip.Debug(message.Fields{
+			"message":  "re-dispatching job that has already been dispatched",
+			"queue_id": d.queue.ID(),
+			"job_id":   job.ID(),
+			"service":  "amboy.dispatcher",
+		})
 		delete(d.cache, job.ID())
 	}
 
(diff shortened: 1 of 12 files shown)


queue/dispatcher.go
```diff
@@ -94,18 +100,35 @@ func (d *dispatcherImpl) Dispatch(ctx context.Context, job amboy.Job) error {
 				return
 			case <-ticker.C:
 				if err := job.Lock(d.queue.ID(), d.queue.Info().LockTimeout); err != nil {
+					grip.Debug(message.WrapError(err, message.Fields{
```
julianedwards, line 103: Should we be nervous about noise caused by all of this logging?

queue/driver_mongo.go
```diff
@@ -880,14 +799,115 @@ RETRY:
...
+			grip.Warning(message.WrapError(err, message.Fields{
+				"id":            d.instanceID,
+				"service":       "amboy.queue.mdb",
+				"operation":     "converting document",
+				"message":       "problem converting job from intermediate form",
+				"is_group":      d.opts.UseGroups,
+				"group":         d.opts.GroupName,
+				"duration_secs": time.Since(startAt).Seconds(),
+			}))
+			// try for the next thing in the iterator if we can
```
julianedwards, line 853: I think you need to set `job = nil` here and before the rest of the `continue` statements.

julianedwards (changes requested):

queue/dispatcher.go
Kimchelly: I wouldn't. The logs don't happen except in cases like server shutdown or somebody manually messing with the job document. The "slow job dispatching operation" already logs hundreds of times per minute.

queue/driver_mongo.go
Kimchelly: We don't because the job is not declared outside the loop.  The only way it could return the job is on L907, after the job is already dispatched.

julianedwards: Ah, right.

julianedwards (approved):

PR #181, opened 2021-02-10
EVG-13540: add SaveAndPut operation

Kimchelly:
Patch: https://evergreen.mongodb.com/version/602454e43627e06ab8f67ebc

~Note: [this PR](https://github.com/mongodb/amboy/pull/180) should be merged first because transactions require MongoDB 4.0 or higher.~

* Add an atomic `SaveAndPut` queue method to update the existing job in the queue and put in a new job (i.e. the retry job). We have to use a transaction to guarantee atomicity of the save/put and also to transfer scopes safely from the old job to the new one (because they have a unique index).
* Split the retryability features into a separate interface (RetryableQueue and RetryableJob) so that queues that don't support  atomic operations can opt out of retryability (e.g. the SQS queue) and to reduce the number of implementations we have to maintain.
* Revendor mongo-driver to include `WithTransaction`.


julianedwards (approved):
I feel like this is going to be tough to test until we release it into the wild. Looks simple enough though, LGTM!

PR #184, opened 2021-02-12
EVG-13543: add test fixtures for retry handler testing

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13543
Patch: https://evergreen.mongodb.com/version/6026e7583627e07a7575d91d

* Add mock implementations for RetryableJob, remoteQueue, and Dispatcher interfaces so that runtime behavior is easy to swap out.
* Add some docs and unit tests for the Dispatcher/remoteQueue.


queue/remote_base.go
```diff
@@ -13,8 +13,13 @@ import (
...
 
+// remoteQueue is an interface to an amboy.RetryableQueue that uses a
+// remoteQueueDriver to interact with the persistence layer for the queue.
 type remoteQueue interface {
 	amboy.RetryableQueue
+	// SetDriver sets the driver to connect to the persistence layer. The driver
+	// must be set before the queue can start. Once the queue has started, the
+	// driver cannot be modified.
 	SetDriver(remoteQueueDriver) error
 	Driver() remoteQueueDriver
```
julianedwards, line 24: Should we add documentation here since it is exported?

queue/dispatcher_test.go
```diff
@@ -0,0 +1,294 @@
...
+	if _, ok := d.dispatched[j.ID()]; ok {
+		return errors.Errorf("cannot dispatch a job more than once")
+	}
+
+	ti := amboy.JobTimeInfo{
+		Start: time.Now(),
+	}
+	j.UpdateTimeInfo(ti)
+
+	if d.initialLock != nil {
```
julianedwards, line 55: What is `initialLock` and should we still call `Save` here?

queue/mock_test.go
```diff
@@ -0,0 +1,289 @@
...
+	completeJob   func(remoteQueue, context.Context, amboy.Job)
+	jobResults    func(remoteQueue, context.Context) <-chan amboy.Job
+	jobStats      func(remoteQueue, context.Context) <-chan amboy.JobStatusInfo
+	queueStats    func(remoteQueue, context.Context) amboy.QueueStats
+	queueInfo     func(remoteQueue) amboy.QueueInfo
+	startQueue    func(remoteQueue, context.Context) error
+	closeQueue    func(remoteQueue, context.Context)
+}
+
+func newMockRemoteQueue(q remoteQueue, driver remoteQueueDriver, makeDispatcher func(q amboy.Queue) Dispatcher, rh amboy.RetryHandler) (*mockRemoteQueue, error) {
```
julianedwards, line 48: Can we use an options struct here?

queue/mock_test.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// mockRetryableQueue provides a mock implementation of a remoteQueue whose
+// runtime behavior is configurable. Methods can be mocked out; otherwise, the
+// default behavior is identical to that of its underlying remoteQueue.
+type mockRemoteQueue struct {
+	remoteQueue
+
+	// Mockable methods
+	putJob        func(remoteQueue, context.Context, amboy.Job) error
```
julianedwards, line 34: Some of these funcs have named arguments and others don't, I think it should be consistent. Also, small nit but can you put `ctx context.Context` as the first arg?

julianedwards (changes requested):

queue/dispatcher_test.go
Kimchelly: `initialLock` is just a mock of taking the distributed locking on the job (i.e. `Lock()` then `Save()`).

queue/mock_test.go
Kimchelly: I only named the ones that had non-obvious arguments (i.e. what does a string mean when passed to Get), but I can name all of them for consistency.

Moved `context.Context`.

queue/mock_test.go
Kimchelly: Done.

queue/remote_base.go
Kimchelly: It's not exported because `remoteQueue` is not exported (I have no idea why), but they could be unexported if we wanted. Even though we don't need them, I added docs for completeness and so that others can understand the interface just from the doc comments.

queue/dispatcher_test.go
julianedwards: Ah, so the `initialLock` function would be expected to call `Save()`. Thanks!

queue/mock_test.go
julianedwards: You don't have to name everything, that is fine. But `context.Context` isn't moved, maybe you forgot to push your changes?

julianedwards (approved):
LGTM modulo `context.Context` as first arg.

queue/mock_test.go
Kimchelly: You're probably viewing the outdated one. The new commit is [this](https://github.com/mongodb/amboy/pull/184/commits/656c7747beadb6887d20bd3e934e5283eec11ea3#diff-3a3bd2a2c3807e9ea4810a5363bc023f9b5441c1f6de21eaf11a3fc0c8a4e3dfR34).

PR #185, opened 2021-02-16
EVG-13540: handle retryable jobs IDs in the driver

Kimchelly:
Patch: https://evergreen.mongodb.com/version/602c406ec9ec445b6f7bcd8c

# Background
Jobs in queue groups are basically the same as regular ungrouped jobs, but they have the group name prepended to them (i.e. the `_id` format is `${groupName}.${jobID}`). The driver deals with adding the group name when it goes into the DB and when queries are made on a queue group, the driver handles re-adding the group name to the query, so that the Amboy end user never has to deal with a grouped job queue differently from a regular job queue.

Since each retry job needs to have a unique `_id`, I originally just glommed the attempt number at the end of the job ID in the retry handler. However, this makes it hard to get jobs out of the driver (e.g. for the Nth job attempt, you would have to call `(amboy.Queue).Get("${myJobID}.attempt-${N}")`. Instead, I changed it so that the driver deals with the `_id` manipulation for retryable jobs. The number is now a prefix and the driver deals with appending/removing the attempt just like how it does for jobs in queue groups. Now `Get("${myJobID}")` will return the latest job attempt for the given job ID.

For a regular job, the `_id` in the DB is still `${myJobID}`.
For a retryable job, the `_id` in the DB is now `attempt-${N}.${myJobID}`.
For a retryable job in a group queue, the `_id` in the DB is now `${groupName}.attempt-${N}.${myJobID}`.

# Changes
* Rename `CurrentTrial` to `CurrentAttempt` to make the naming consistent.
* Move retry `_id` handling logic out of the retry handler and into the driver.
* `Get()` works the same as before. If it's passed a job ID for a retryable job, it returns the latest job attempt.
* Add new `GetAttempt()` to get the Nth attempt of a job rather than the latest.
* Rename some group-handling helper functions and combine them with retry attempt-handling helper functions.
* Remove some places where the group is unnecessarily removed from the job interchange name (it's not necessarily wrong, but it doesn't do anything so I would prefer to simplify it).
* Do some variable renaming in the driver for clarification, because `j` and `job` are used inconsistently between `Job`s and `JobInterchange`s.

### DIFF
--- a/queue/driver.go
+++ b/queue/driver.go
@@ -53,7 +53,6 @@ type MongoDBOptions struct {
 	DB                       string
 	GroupName                string
 	UseGroups                bool
-	UseRetries               bool
 	Priority                 bool
 	CheckWaitUntil           bool
 	CheckDispatchBy          bool
--- a/queue/driver.go
+++ b/queue/driver.go
@@ -15,8 +15,13 @@ type remoteQueueDriver interface {
 	Open(context.Context) error
 	Close()
 
-	// Get finds a job by job ID.
+	// Get finds a job by job ID. For retryable jobs, this returns the latest
+	// job attempt.
 	Get(context.Context, string) (amboy.Job, error)
+	// GetAttempt returns a job by job ID and attempt number. Only applicable to
+	// retryable jobs. If used for a non-retryable job, this should return nil
+	// and an error.
+	GetAttempt(ctx context.Context, id string, attempt int) (amboy.RetryableJob, error)
 	// Put inserts a new job in the backing storage.
 	Put(context.Context, amboy.Job) error
 	// Save updates an existing job in the backing storage. Implementations may
@@ -48,6 +53,7 @@ type MongoDBOptions struct {
 	DB                       string
 	GroupName                string
 	UseGroups                bool
+	UseRetries               bool
 	Priority                 bool
 	CheckWaitUntil           bool
 	CheckDispatchBy          bool
--- a/queue/driver_mongo.go
+++ b/queue/driver_mongo.go
@@ -238,6 +238,19 @@ func (d *mongoDriver) queueIndexes() []mongo.IndexModel {
 			},
 			Options: options.Index().SetSparse(true).SetUnique(true),
 		},
+		{
+			Keys: bsonx.Doc{
+				{
+					Key:   "retry_info.base_job_id",
+					Value: bsonx.Int32(1),
+				},
+				{
+					Key:   "retry_info.current_attempt",
+					Value: bsonx.Int32(-1),
+				},
+			},
+			Options: options.Index().SetUnique(true),
+		},
 	}
 
 	if d.opts.TTL > 0 {
@@ -250,22 +263,6 @@ func (d *mongoDriver) queueIndexes() []mongo.IndexModel {
 				},
 			},
 			Options: options.Index().SetExpireAfterSeconds(ttl),
-		})
-	}
-
-	if d.opts.UseRetries {
-		indexes = append(indexes, mongo.IndexModel{
-			Keys: bsonx.Doc{
-				{
-					Key:   "retry_info.base_job_id",
-					Value: bsonx.Int32(1),
-				},
-				{
-					Key:   "retry_info.current_attempt",
-					Value: bsonx.Int32(-1),
-				},
-			},
-			Options: options.Index().SetUnique(true),
 		})
 	}
 
--- a/queue/driver_mongo.go
+++ b/queue/driver_mongo.go
@@ -400,9 +400,6 @@ func (d *mongoDriver) getIDWithGroup(name string) string {
 }
 
 func (d *mongoDriver) addMetadata(j *registry.JobInterchange) {
-	if j.RetryInfo.Retryable {
-		j.RetryInfo.BaseJobID = j.Name
-	}
 	d.addRetryToMetadata(j)
 	d.addGroupToMetadata(j)
 }
@@ -434,6 +431,7 @@ func (d *mongoDriver) addRetryToMetadata(j *registry.JobInterchange) {
 		return
 	}
 
+	j.RetryInfo.BaseJobID = j.Name
 	j.Name = buildCompoundID(retryAttemptPrefix(j.RetryInfo.CurrentAttempt), j.Name)
 }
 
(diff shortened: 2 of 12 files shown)


queue/driver_mongo.go
```diff
@@ -419,8 +474,6 @@ func (d *mongoDriver) Get(ctx context.Context, name string) (amboy.Job, error) {
 		return nil, errors.Wrapf(err, "GET problem decoding '%s'", name)
 	}
 
-	d.processNameForUsers(j)
```
julianedwards, line 422: Should this call `d.removeMetadata()`?

queue/driver_mongo.go
```diff
@@ -430,19 +483,48 @@ func (d *mongoDriver) Get(ctx context.Context, name string) (amboy.Job, error) {
...
+	res := d.getCollection().FindOne(ctx, matchIDAndAttempt)
+	if err := res.Err(); err != nil {
+		return nil, errors.Wrapf(err, "GET problem fetching '%s'", name)
+	}
+
+	ji := &registry.JobInterchange{}
+	if err := res.Decode(ji); err != nil {
+		return nil, errors.Wrapf(err, "GET problem decoding '%s'", name)
+	}
+
```
julianedwards, line 500: Should this call `d.removeMetadata`?

queue/driver.go
```diff
@@ -15,8 +15,13 @@ type remoteQueueDriver interface {
 	Open(context.Context) error
 	Close()
 
-	// Get finds a job by job ID.
+	// Get finds a job by job ID. For retryable jobs, this returns the latest
+	// job attempt.
```
julianedwards, line 19: Would it make sense to have another `Get` that gets the latest job attempt and returns an `amboy.RetryableJob`. More of a convenience function than anything, but would reduce all the type checking.

julianedwards (changes requested):

queue/driver.go
Kimchelly: I feel like that would be annoying because you can't overload method names unless I made a `GetLatest` on the `RetryableQueue` which behaves identically to `Get()`. There's only one place that depends on `Get()` returning a retryable job.

queue/driver_mongo.go
```diff
@@ -374,23 +387,32 @@ func (d *mongoDriver) Close() {
...
 }
 
-func (d *mongoDriver) processNameForUsers(j *registry.JobInterchange) {
-	if !d.opts.UseGroups {
-		return
+func (d *mongoDriver) addMetadata(j *registry.JobInterchange) {
+	if j.RetryInfo.Retryable {
+		j.RetryInfo.BaseJobID = j.Name
 	}
+	d.addRetryToMetadata(j)
```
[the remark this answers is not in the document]
Kimchelly: Moved it into `addRetryToMetadta()`.

queue/driver_mongo.go
```diff
@@ -399,6 +421,30 @@ func (d *mongoDriver) processJobForGroup(j *registry.JobInterchange) {
...
+
+	j.Name = buildCompoundID(retryAttemptPrefix(j.RetryInfo.CurrentAttempt), j.Name)
+}
+
+func (d *mongoDriver) removeRetryFromMetadata(j *registry.JobInterchange) {
+	if !j.RetryInfo.Retryable {
+		return
+	}
+
+	j.Name = deconstructCompoundID(j.Name, retryAttemptPrefix(j.RetryInfo.CurrentAttempt))
```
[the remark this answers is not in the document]
Kimchelly: This shouldn't be possible because `removeMetadata()` is only used on jobs deserialized from the DB, which means that the only way someone could break this would be by changing the driver to modify the attempt. This method is only necessary for `JobStats` to return the job ID without the retry/group info.

queue/driver_mongo.go
```diff
@@ -410,7 +456,16 @@ func (d *mongoDriver) modifyQueryForGroup(q bson.M) {
 func (d *mongoDriver) Get(ctx context.Context, name string) (amboy.Job, error) {
 	j := &registry.JobInterchange{}
 
-	res := d.getCollection().FindOne(ctx, bson.M{"_id": d.getIDFromName(name)})
+	matchID := bson.M{
+		"$or": []bson.M{
+			{"_id": d.getIDWithGroup(name)},
+			{"retry_info.base_job_id": name},
```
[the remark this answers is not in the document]
Kimchelly: It's not supposed to have the group name because users aren't supposed to be aware of the format of the stored `_id`, only the logical job ID. I purposely wrote `addMetadata` and `removeMetadata` so that when jobs go into the DB, the base job ID is set before changing the `_id`.

Putting the same job in two different group queues would be invalid. I don't see a good use case for putting the same job in two logically separate queues (and typically it just leads to bugs), because it'll just cause your job to run twice instead of once.

Kimchelly:
Patch: https://evergreen.mongodb.com/version/602d89b31e2d17090b720fb4

queue/driver.go
julianedwards: Sounds good.

queue/driver_mongo.go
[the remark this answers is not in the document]
julianedwards: I think my confusion is that this is an `$or`, so it seems like group name for retryable jobs is pointless? I guess I am confused on the point of prepending the group name then at all.

queue/driver_mongo.go
julianedwards: ?

queue/driver_mongo.go
julianedwards: ?

julianedwards (changes requested):
I think you missed a few of my comments.

queue/driver_mongo.go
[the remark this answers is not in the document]
Kimchelly: The group name is not used in base job ID. The group name prepending is the current behavior - it has to be prepended for backward compatibility.

queue/driver_mongo.go
Kimchelly: I mentioned this in the original PR comment but this code is not necessary. All it does is modify the job interchange name, which is not necessary if all you want is to return the Job object.

queue/driver_mongo.go
Kimchelly: Same thing as above - since this is only used to return the Job object, it doesn't need to JobInterchange name at all.

julianedwards (approved):

PR #188, opened 2021-02-18
EVG-13540: add unit and integration tests for retry handler

Kimchelly:

* Add unit testing and integration testing for the retry handler.
* Make some minor fixes to the retry handler around locking, restoring in-memory job state, and loading the job from the DB.

### DIFF
--- a/queue/mock_test.go
+++ b/queue/mock_test.go
@@ -103,8 +103,10 @@ func (q *mockRemoteQueue) Driver() remoteQueueDriver {
 }
 
 func (q *mockRemoteQueue) SetDriver(d remoteQueueDriver) error {
+	if err := q.remoteQueue.SetDriver(d); err != nil {
+		return err
+	}
 	q.driver = d
-	q.remoteQueue.SetDriver(d)
 	q.driver.SetDispatcher(q.dispatcher)
 	return nil
 }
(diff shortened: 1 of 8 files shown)


queue/mock_test.go
```diff
@@ -66,25 +71,51 @@ func newMockRemoteQueue(opts mockRemoteQueueOptions) (*mockRemoteQueue, error) {
...
 	mq := &mockRemoteQueue{remoteQueue: opts.queue}
 	if opts.driver != nil {
-		if err := opts.queue.SetDriver(opts.driver); err != nil {
+		if err := mq.SetDriver(opts.driver); err != nil {
 			return nil, errors.Wrap(err, "configuring queue with driver")
 		}
 	}
 	if opts.makeDispatcher != nil {
-		dispatcher := opts.makeDispatcher(mq)
+		mq.dispatcher = opts.makeDispatcher(mq)
```
julianedwards, line 79: Does it make sense to make this even if there is not driver? In other words, should the above line be `if opts.makeDispatcher != nil && opts.driver != nil`?

julianedwards (approved):
LGTM modulo nit

queue/mock_test.go
Kimchelly: I set the mock queue to initialize a default dispatcher if it's unset, since the driver can't realistically operate without a dispatcher.

PR #189, opened 2021-02-18
EVG-13540: make retry handler more robust

Kimchelly:
Patch: https://evergreen.mongodb.com/version/602fe3331e2d171e697ea37a

* Add configurable wait interval between checking for new jobs.
* Safely unset the scopes the old job was holding onto in the retry handler and change the method to `CompleteAndPut()`. This is to handle the case in which the old job is trying to atomically hand the scopes to the new job. The job has to hold them until the retry handler finishes processing the job, after which point it should release the scopes.
* Try multiple times to mark the job as processed using `CompleteRetry()`. [EVG-13784](https://jira.mongodb.org/browse/EVG-13784) will deal with the edge case of  it failing to save the job by ensuring the job does eventually unset NeedsRetry and its scopes (e.g. most likely, either the context errored or the job document was changed so that the in-memory copy is outdated). That ticket will ensure the job does not just retry infinitely and does not hold its scopes forever.

### DIFF
--- a/queue/driver_mongo.go
+++ b/queue/driver_mongo.go
@@ -612,17 +612,6 @@ func (d *mongoDriver) SaveAndPut(ctx context.Context, toSave amboy.Job, toPut am
 	defer sess.EndSession(ctx)
 
 	atomicSaveAndPut := func(sessCtx mongo.SessionContext) (interface{}, error) {
-		// TODO (EVG-13540): for a long-running job that has been lock pinged,
-		// we need to make sure that the modTS/modcount is okay to use. Save()
-		// should still succeed as long as:
-		// * Continuity of in-memory job: the dispatcher shares the same
-		// in-memory copy of the job as the retry handler.
-		// * Concurrency: the dispatcher never runs concurrently with the retry
-		// handler (which would modify modTS/modcount).
-		// * Precondition: the job is complete when this is called.
-		// We should verify these statements are true when doing tests of the
-		// retry handler.
-
 		if err = d.Save(sessCtx, toSave); err != nil {
 			return nil, errors.Wrap(err, "saving old job")
 		}
--- a/queue/driver_mongo.go
+++ b/queue/driver_mongo.go
@@ -604,15 +604,15 @@ func (d *mongoDriver) Save(ctx context.Context, j amboy.Job) error {
 	return errors.WithStack(d.doUpdate(ctx, ji))
 }
 
-func (d *mongoDriver) SaveAndPut(ctx context.Context, toSave amboy.Job, toPut amboy.Job) error {
+func (d *mongoDriver) CompleteAndPut(ctx context.Context, toComplete amboy.Job, toPut amboy.Job) error {
 	sess, err := d.client.StartSession()
 	if err != nil {
 		return errors.Wrap(err, "starting transaction session")
 	}
 	defer sess.EndSession(ctx)
 
-	atomicSaveAndPut := func(sessCtx mongo.SessionContext) (interface{}, error) {
-		if err = d.Complete(sessCtx, toSave); err != nil {
+	atomicCompleteAndPut := func(sessCtx mongo.SessionContext) (interface{}, error) {
+		if err = d.Complete(sessCtx, toComplete); err != nil {
 			return nil, errors.Wrap(err, "completing old job")
 		}
 
@@ -623,7 +623,7 @@ func (d *mongoDriver) SaveAndPut(ctx context.Context, toSave amboy.Job, toPut am
 		return nil, nil
 	}
 
-	if _, err = sess.WithTransaction(ctx, atomicSaveAndPut); err != nil {
+	if _, err = sess.WithTransaction(ctx, atomicCompleteAndPut); err != nil {
 		return errors.Wrap(err, "atomic save and put")
 	}
 
--- a/queue/driver_mongo.go
+++ b/queue/driver_mongo.go
@@ -624,7 +624,7 @@ func (d *mongoDriver) CompleteAndPut(ctx context.Context, toComplete amboy.Job, 
 	}
 
 	if _, err = sess.WithTransaction(ctx, atomicCompleteAndPut); err != nil {
-		return errors.Wrap(err, "atomic save and put")
+		return errors.Wrap(err, "atomic complete and put")
 	}
 
 	return nil
--- a/queue/driver_mongo.go
+++ b/queue/driver_mongo.go
@@ -594,14 +594,14 @@ func isMongoDupKey(err error) bool {
 }
 
 func (d *mongoDriver) Save(ctx context.Context, j amboy.Job) error {
-	job, err := d.prepareInterchange(j)
+	ji, err := d.prepareInterchange(j)
 	if err != nil {
 		return errors.WithStack(err)
 	}
 
-	job.Scopes = j.Scopes()
+	ji.Scopes = j.Scopes()
 
-	return errors.WithStack(d.doUpdate(ctx, job))
+	return errors.WithStack(d.doUpdate(ctx, ji))
 }
 
 func (d *mongoDriver) SaveAndPut(ctx context.Context, toSave amboy.Job, toPut amboy.Job) error {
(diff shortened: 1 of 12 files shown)


queue/driver_mongo.go
```diff
@@ -594,37 +594,26 @@ func isMongoDupKey(err error) bool {
...
-		// * Concurrency: the dispatcher never runs concurrently with the retry
-		// handler (which would modify modTS/modcount).
-		// * Precondition: the job is complete when this is called.
-		// We should verify these statements are true when doing tests of the
-		// retry handler.
-
-		if err = d.Save(sessCtx, toSave); err != nil {
-			return nil, errors.Wrap(err, "saving old job")
+	atomicCompleteAndPut := func(sessCtx mongo.SessionContext) (interface{}, error) {
+		if err = d.Complete(sessCtx, toComplete); err != nil {
```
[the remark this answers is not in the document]
Kimchelly: The TODO is a bit outdated. I don't think this will be a concern because the dispatcher is already done by the time this code is reached and all threads hold onto the same in-memory job from the moment it leaves the queue to the moment the retry handler receives it. Any other concurrency concerns between retry handlers on different hosts will be handled by [EVG-13784](https://jira.mongodb.org/browse/EVG-13784).

queue/driver_mongo.go
```diff
@@ -634,35 +623,28 @@ func (d *mongoDriver) SaveAndPut(ctx context.Context, toSave amboy.Job, toPut am
...
-		job.Scopes = nil
-	})
-
-	if !isRetryableJob {
-		job.Scopes = nil
+	// It is safe to drop the scopes now in all cases except for one - if the
+	// job still needs to retry and applies its scopes immediately to the retry
+	// job, we cannot let go of the scopes yet because they will need to be
+	// safely transferred to the retry job.
+	if !ji.RetryInfo.Retryable || !ji.RetryInfo.NeedsRetry || !ji.ApplyScopesOnEnqueue {
```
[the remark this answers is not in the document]
Kimchelly: For scopes applied on enqueue, the hand-off has to occur in the retry handler, so it's deferring the scope lock drop until `CompleteAndPut` occurs.

This logic is identical to what it was before, but I applied DeMorgan's law to make the conditional logic less awkward-looking. It's equivalent to this:
```go
if ji.RetryInfo.Retryable && ji.RetryInfo.NeedsRetry && ji.ApplyScopesOnEnqueue {
    // pass, hold the scope
} else {
    ji.Scopes = nil
 }
```

queue/remote_base.go
```diff
@@ -248,6 +249,45 @@ func (q *remoteBase) Complete(ctx context.Context, j amboy.Job) {
...
+			catcher.Add(ctx.Err())
+			return errors.Wrapf(catcher.Resolve(), "giving up after attempt %d", attempt)
+		case <-timer.C:
+			j.UpdateRetryInfo(amboy.JobRetryOptions{
+				NeedsRetry: utility.FalsePtr(),
+			})
+
+			if err := q.driver.Complete(ctx, j); err != nil {
+				catcher.Wrapf(err, "attempt %d", attempt)
+				if attempt >= 10 || isMongoDupKey(err) {
```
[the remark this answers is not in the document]
Kimchelly: True, I just deleted the attempt check since it'll be checked by the for loop. This is what I get for copy-pasting from the `Complete()` code.

queue/retry.go
```diff
@@ -154,20 +154,18 @@ func (rh *basicRetryHandler) waitForJob(ctx context.Context) error {
 					"queue_id": rh.queue.ID(),
 					"job_id":   j.ID(),
 				}))
+				j.AddError(err)
+
+				if err := rh.queue.CompleteRetry(ctx, j); err != nil {
```
julianedwards, line 159: Can we add a comment here that we are calling `CompleteRetry` to ensure the job is marked as complete since `CompleteAndPut` failed in `handleJob`?

queue/driver_mongo.go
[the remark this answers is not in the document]
julianedwards: Maybe I am missing something, but I do not see where the scope hand-off is happening in the retry handler. I am assuming it would be in `tryEnqueueJob` after `CompleteAndPut` is called. Could you point me to it?

queue/remote_base.go
```diff
@@ -248,6 +249,46 @@ func (q *remoteBase) Complete(ctx context.Context, j amboy.Job) {
...
+		case <-timer.C:
+			j.UpdateRetryInfo(amboy.JobRetryOptions{
+				NeedsRetry: utility.FalsePtr(),
+			})
+
+			if err := q.driver.Complete(ctx, j); err != nil {
+				catcher.Wrapf(err, "attempt %d", attempt)
+				if isMongoDupKey(err) {
+					j.AddError(catcher.Resolve())
+					return errors.Wrapf(catcher.Resolve(), "giving up after attempt %d", attempt)
```
julianedwards, line 278: Nit: you can just break here and let the `return` on line 289 take care of this.

queue/remote_base.go
```diff
@@ -248,6 +249,46 @@ func (q *remoteBase) Complete(ctx context.Context, j amboy.Job) {
...
+	defer timer.Stop()
+
+	var attempt int
+	catcher := grip.NewBasicCatcher()
+
+	for attempt = 1; attempt <= 10; attempt++ {
+		select {
+		case <-ctx.Done():
+			catcher.Add(ctx.Err())
+			return errors.Wrapf(catcher.Resolve(), "giving up after attempt %d", attempt)
```
julianedwards, line 268: Nit: you can just break here and let the `return` on line 289 take care of this.

queue/remote_base.go
```diff
@@ -248,6 +249,46 @@ func (q *remoteBase) Complete(ctx context.Context, j amboy.Job) {
...
+				if isMongoDupKey(err) {
+					j.AddError(catcher.Resolve())
+					return errors.Wrapf(catcher.Resolve(), "giving up after attempt %d", attempt)
+				}
+
+				timer.Reset(retryInterval)
+				continue
+			}
+
+			return nil
```
julianedwards, line 285: I think you can remove this, otherwise you can never reach the return outside of the for loop.

queue/driver_mongo.go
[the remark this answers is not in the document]
Kimchelly: It's implicit by calling `GetAttempt()` since it receives an in-memory copy of the scopes it should acquire. Dropping the scopes here by settings `NeedsRetry = false` is the only step necessary for the new job to acquire them in `Put()` (L531).

queue/remote_base.go
Kimchelly: Since it's equivalent, I would prefer to just not break if it's not actually necessary to make it more obvious that it should give up on context error.

queue/remote_base.go
Kimchelly: Same thing as above.

queue/remote_base.go
Kimchelly: This is the success case, so it would be incorrect to return an error here.

queue/retry.go
Kimchelly: Done.

queue/retry.go
```diff
@@ -154,20 +154,18 @@ func (rh *basicRetryHandler) waitForJob(ctx context.Context) error {
...
-			j.UpdateRetryInfo(amboy.JobRetryOptions{
-				NeedsRetry: utility.FalsePtr(),
-			})
-			// TODO (EVG-13540): this has to retry this op until success,
-			// like the theoretical infinite loop in queue.Complete(). It should
-			// also be done in a transaction-like way when the new job is
-			// inserted, so that the swap occurs safely.
-			if err := rh.queue.Save(ctx, j); err != nil {
-				grip.Critical(message.WrapError(err, message.Fields{}))
+			if err := rh.queue.CompleteRetry(ctx, j); err != nil {
```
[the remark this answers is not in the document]
Kimchelly: Because otherwise, the job will retry again since NeedsRetry = true still and we need a means to abort and never retry this job again, since it's unable to retry for whatever reason.

queue/remote_base.go
julianedwards: Oh right, my bad. Cause you have the `continue` on L281.

queue/driver_mongo.go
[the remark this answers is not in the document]
julianedwards: Ah ok, got it.

julianedwards (approved):

PR #190, opened 2021-02-22
EVG-13591: add error type for duplicate scopes

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13591
Patch: https://evergreen.mongodb.com/version/603437170305b916143db426

* Add a new error type, which is a type of duplicate key error, specifically for handling duplicate key errors due to another job holding the scope. For `amboy.EnqueueUniqueJob()`, it'll still squash errors due to acquiring duplicate scopes. If you want to check specifically for the duplicate scope case, you can just due `(amboy.Queue).Put()` and check for the duplicate key or scope error. The check is kind of fragile since it depends on parsing the error message text, but it seems like the server doesn't provide a nicer way to determine the index that caused the write failure.
* Return duplicate job scope errors if insertion fails because of the uniqueness constraint of scopes.

### DIFF
--- a/queue/driver_mongo.go
+++ b/queue/driver_mongo.go
@@ -535,6 +535,9 @@ func (d *mongoDriver) Put(ctx context.Context, j amboy.Job) error {
 
 	if _, err = d.getCollection().InsertOne(ctx, ji); err != nil {
 		if isMongoDupKey(err) {
+			if isMongoDupScope(err) {
+				return amboy.NewDuplicateJobScopeErrorf("job scopes '%s' conflict", j.Scopes())
+			}
 			return amboy.NewDuplicateJobErrorf("job '%s' already exists", j.ID())
 		}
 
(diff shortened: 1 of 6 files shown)


queue/driver_mongo.go
```diff
@@ -573,26 +576,102 @@ func (d *mongoDriver) getAtomicQuery(jobName string, modCount int) bson.M {
...
+}
+
+func getMongoDupKeyWriteConcernError(err mongo.WriteException) *mongo.WriteConcernError {
+	wce := err.WriteConcernError
+	if wce == nil {
+		return nil
+	}
+
+	switch wce.Code {
+	case 11000, 11001, 12582:
```
julianedwards, line 634: Can we put comments on what these codes mean and why we ignore some?

julianedwards (changes requested):

queue/driver_mongo.go
Kimchelly: This logic has existed already, I just moved it into a switch. This appears to be [copy-pasted from the go driver](https://github.com/mongodb/mongo-go-driver/blob/213fb80b373f70dba4f9f516dc4c718abe41c76b/mongo/errors.go#L87-L96). When we upgrade the go driver to the latest, we could potentially replace this.

julianedwards: Is there a reason we can't update the go driver now?

Kimchelly: Mostly I don't want to deal with the possibility that it doesn't work for some subtle reason and also that I don't want to revendor the driver in Barque/Cedar/Evergreen just for a tiny piece of code. I also have some concerns that it might not work the same as it did before because the duplicate key error checks differ between server versions. We can deal with upgrading the Go driver to the latest version later.

julianedwards (approved):

PR #191, opened 2021-02-22
EVG-13544: add option for max job attempts

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13544
Patch: https://evergreen.mongodb.com/version/60341d202a60ed7e784ec73f

Add job option to put an upper limit on the number of times it will retry.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -170,17 +170,32 @@ type JobRetryInfo struct {
 	// (unless otherwise set on enqueue), so the first time the job attempts to
 	// run, its value is 0. Each subsequent retry increments this value.
 	CurrentAttempt int `bson:"current_attempt" json:"current_attempt,omitempty" yaml:"current_attempt,omitempty"`
+	// MaxAttempts is the maximum number of attempts for a job. This is
+	// 1-indexed since it is a count. For example, if this is set to 3, the job
+	// will be allowed to run 3 times at most. If unset, the default maximum
+	// attempts is 10.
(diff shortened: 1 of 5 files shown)


interface.go
```diff
@@ -366,7 +382,9 @@ type RetryHandler interface {
 // RetryHandlerOptions configures the behavior of a RetryHandler.
 type RetryHandlerOptions struct {
 	// MaxRetryAttempts is the maximum number of times that the retry handler is
-	// allowed to attempt to retry a job before it gives up.
+	// allowed to attempt to retry a job before it gives up. This is referring
+	// to the retry handler's attempts to internally retry the job and is
+	// unrelated to the job's particular max attempt setting.
```
julianedwards, line 387: I am assuming this takes precedence over a job's `MaxAttempts` setting. If so, can we specify that here?

julianedwards (changes requested):

interface.go
Kimchelly: I'm not sure what you mean. This is the job-specific max attempts. If you're talking about the retry handler's max attempts, that's referring to the number of times the retry handler is allowed to attempt to re-enqueue the job before it gives up on it and says it's unretryable.

julianedwards: Oh, ok I see now. So the job's attempt count will not actually increase until it is enqueued?

julianedwards (approved):

interface.go
Kimchelly: Yes, this is referring to the `CurrentAttempt` persisted on the job. The retry handler's count just an in-memory count of how many times it's tried to handle the job.

PR #192, opened 2021-02-23
EVG-13784: monitor stale retrying jobs

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13784

* Add a goroutine to monitor for job that are in the process of retrying but are not actively marked as retrying (i.e. the lock timeout is exceeded).
    * Add a driver method to return retryable jobs based on a filter.
* Atomically lock a retry job to a single in-memory queue's retry handler.
* Move a few random helper functions to more appropriate places.

PR #193, opened 2021-02-24
EVG-13541, EVG-13542: add options for DispatchBy and WaitFor

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13541
https://jira.mongodb.org/browse/EVG-13542
Patch: https://evergreen.mongodb.com/version/60366e44a4cf47157e163468

* Add settings to modify DispatchBy and WaitFor on retried jobs.
* Rename functions (i.e. CompleteRetry -> CompleteRetrying and CompleteAndPut -> CompleteRetryingAndPut) to make it more explicit that this is marking a retrying job as processed (instead of marking a regular job as completed executing).
* Push job metadata updates into the queue level to be consistent with guarantees on other queue methods (e.g. Put must set the job creation time).

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -366,15 +366,14 @@ type RetryableQueue interface {
 	// and a bool indicating whether the job was found or not.
 	GetAttempt(ctx context.Context, id string, attempt int) (RetryableJob, bool)
 
-	// CompleteAndPut marks an existing job toComplete in the queue (see
-	// CompleteRetry) as finished processing its retry and inserts a new job
-	// toPut in the queue (see Put). Implementations must make this operation
-	// atomic.
-	CompleteAndPut(ctx context.Context, toComplete, toPut Job) error
+	// CompleteRetryingAndPut marks an existing job toComplete in the queue (see
+	// CompleteRetrying) as finished retrying and inserts a new job toPut in the
+	// queue (see Put). Implementations must make this operation atomic.
+	CompleteRetryingAndPut(ctx context.Context, toComplete, toPut RetryableJob) error
 
-	// CompleteRetry marks a job that needs to retry as finished processing, so
+	// CompleteRetrying marks a job that is retrying as finished processing, so
 	// that it will no longer retry.
-	CompleteRetry(ctx context.Context, j RetryableJob) error
+	CompleteRetrying(ctx context.Context, j RetryableJob) error
 }
 
 // RetryHandler provides a means to retry RetryableJobs within a RetryableQueue.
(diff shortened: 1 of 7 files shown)


interface.go
```diff
@@ -352,15 +366,14 @@ type RetryableQueue interface {
...
+	// CompleteRetryingAndPut marks an existing job toComplete in the queue (see
+	// CompleteRetrying) as finished retrying and inserts a new job toPut in the
+	// queue (see Put). Implementations must make this operation atomic.
+	CompleteRetryingAndPut(ctx context.Context, toComplete, toPut RetryableJob) error
 
-	// CompleteRetry marks a job that needs to retry as finished processing, so
+	// CompleteRetrying marks a job that is retrying as finished processing, so
 	// that it will no longer retry.
-	CompleteRetry(ctx context.Context, j RetryableJob) error
+	CompleteRetrying(ctx context.Context, j RetryableJob) error
```
julianedwards, line 376: I thought `CompleteRetry` was clear and bit more concise, but I am not too opinionated on this change.

interface.go
```diff
@@ -175,6 +175,16 @@ type JobRetryInfo struct {
 	// will be allowed to run 3 times at most. If unset, the default maximum
 	// attempts is 10.
 	MaxAttempts int `bson:"max_attempts,omitempty" json:"max_attempts,omitempty" yaml:"max_attempts,omitempty"`
+	// DispatchBy reflects the amount of time (relative to when the job is
+	// retried) that the retried job has to dispatch for execution. If this
+	// deadline elapses, the job will not run. This is analogous to
+	// (JobTimeInfo).DispatchBy.
+	DispatchBy time.Duration `bson:"dispatch_by,omitempty" json:"dispatch_by,omitempty" yaml:"dispatch_by,omitempty"`
```
julianedwards, line 182: So the retry job will not inherit the `DispatchBy` and `WaitUntil` settings of the original job?

Kimchelly: It inherits the original value from the job if it's not specified here.

interface.go
Kimchelly: I mostly did this to be clearer and consistent with the naming of each state. In the [other PR](https://github.com/mongodb/amboy/pull/192), I referred to jobs that are currently in the process of retrying as "retrying" to distinguish this status from just regular "retryable" (but not actively retrying) jobs. I think I would prefer to keep it like this to reduce ambiguity about what state transition it's used for. Without specifying that the job is retrying, it's unclear whether it means "complete executing a retryable job" (not what it means, retryable jobs still call `Complete()` like normal jobs) versus "marking a job as no longer retrying".

julianedwards (approved):

PR #194, opened 2021-02-25
EVG-13784: add retry handler and remote queue integration tests

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13784
Patch: https://evergreen.mongodb.com/version/6037d7ae3e8e8669cc1ed35f

* Add integration tests for retry handler and remote queue.
* Add temporary internal interface for checking runtime stats of retry handler. I might get rid of it after [EVG-14111](https://jira.mongodb.org/browse/EVG-14111), since I mostly wanted it for more convenient testing.
* Reduce the amount of duplicate smok tests run.

### DIFF
--- a/queue/queue_test.go
+++ b/queue/queue_test.go
@@ -411,7 +411,14 @@ func TestQueueSmoke(t *testing.T) {
 				}
 
 				t.Run(runner.Name+"Pool", func(t *testing.T) {
-					var testRetryOnce, testWaitUntilOnce, testDispatchBeforeOnce, testMaxTimeOnce, testScopesOnce, testApplyScopesOnEnqueueOnce sync.Once
+					var (
+						testRetryOnce                sync.Once
+						testWaitUntilOnce            sync.Once
+						testDispatchBeforeOnce       sync.Once
+						testMaxTimeOnce              sync.Once
+						testScopesOnce               sync.Once
+						testApplyScopesOnEnqueueOnce sync.Once
+					)
 
 					for _, size := range DefaultSizeTestCases() {
 						if test.MaxSize > 0 && size.Size > test.MaxSize {
--- a/queue/queue_test.go
+++ b/queue/queue_test.go
@@ -175,7 +175,7 @@ func MongoDBQueueTestCases(client *mongo.Client) []QueueTestCase {
 					Size:    size,
 					Name:    name,
 					Ordered: false,
-					MDB:     DefaultMongoDBOptions(),
+					MDB:     defaultMongoDBTestOptions(),
 					Client:  client,
 				}
 				opts.MDB.Format = amboy.BSON2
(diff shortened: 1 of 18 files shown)


queue/queue_test.go
```diff
@@ -411,6 +411,8 @@ func TestQueueSmoke(t *testing.T) {
 				}
 
 				t.Run(runner.Name+"Pool", func(t *testing.T) {
+					var testRetryOnce, testWaitUntilOnce, testDispatchBeforeOnce, testMaxTimeOnce, testScopesOnce, testApplyScopesOnEnqueueOnce sync.Once
```
julianedwards, line 414: nit: can you make this more readable by doing:
```
var (
        testRetryOnce sync.Once
        testWaitUntilOnce
        ...
)
```

queue/retry.go
```diff
@@ -178,6 +204,13 @@ func (rh *basicRetryHandler) waitForJob(ctx context.Context) {
 				}
 			}
 
+			rh.mu.Lock()
```
julianedwards, line 207: Should L177 also acquire the `rh` mutex when putting the job in the `pending` map?

queue/retry.go
```diff
@@ -11,12 +11,36 @@ import (
...
+type retryingJobOutcome struct {
+	job     amboy.RetryableJob
+	retried bool
+}
+
 type basicRetryHandler struct {
 	queue         amboy.RetryableQueue
 	opts          amboy.RetryHandlerOptions
 	pending       map[string]amboy.RetryableJob
+	completed     map[string]retryingJobOutcome
```
julianedwards, line 43: Will holding pending and completed jobs in memory cause issues?

queue/retry.go
```diff
@@ -11,12 +11,36 @@ import (
...
+	// stats returns information about the job statistics for an
+	// amboy.RetryHandler. Results are reset with each runtime execution.
+	stats() retryHandlerStats
+}
+
+// retryHandlerStats captures runtime statistics on an amboy.RetryHandler.
+type retryHandlerStats struct {
+	pending   int
+	completed int
+	retried   int
```
julianedwards, line 31: Can we call this something else to make this more obvious that it is an aggregated count of errored completed retry jobs? Maybe it is just me, but I think at first glance `retried` isn't obvious. I didn't really get what the stat was telling us until I read L210.

julianedwards (changes requested):

queue/queue_test.go
Kimchelly: Done.

queue/retry.go
Kimchelly: These are temporary stats until [EVG-14111](https://jira.mongodb.org/browse/EVG-14111) is done, so it doesn't seem worth it.

queue/retry.go
Kimchelly: Same comment about temporary stats.

queue/retry.go
Kimchelly: Same comment about temporary stats.

queue/retry.go
julianedwards: SGTM

julianedwards (approved):

PR #195, opened 2021-02-26
EVG-14068: include queue stats for number of retrying jobs

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-14068
Patch: https://evergreen.mongodb.com/version/60393357c9ec44288d41a5ff

* Queues aren't considered complete unless jobs are also done retrying. I changed the statistics counting to an aggregation to reduce the amount of inconsistency in the queue statistics (because it issued 1 query per statistic). There's still a small amount of inconsistency around this because the completed jobs count is indirectly calculated using two non-atomic queries (total - pending), but I think it's only a minor correctness issue.

* Add statistics for number of retrying jobs.
* Add indexes for retrying jobs.
* Clean up some of the driver queries.
* Add warning about pruning queues based on number of completed jobs.


queue/driver_mongo.go
```diff
@@ -1260,58 +1250,139 @@ func (d *mongoDriver) scopesInUse(ctx context.Context, scopes []string) bool {
...
+	q["status.in_prog"] = true
+	return q
+}
+
+func (d *mongoDriver) getCompletedQuery(q bson.M) bson.M {
+	q["status.completed"] = true
+	return q
+}
+
+// getPendingQuery modifies the query to find jobs that are retrying.
```
julianedwards, line 1378: `getRetryingQuery`

queue/driver_mongo.go
```diff
@@ -1260,58 +1250,139 @@ func (d *mongoDriver) scopesInUse(ctx context.Context, scopes []string) bool {
...
+}
+
+// getInProgQuery modifies the query to find jobs that are in progress.
+func (d *mongoDriver) getInProgQuery(q bson.M) bson.M {
+	q = d.getPendingQuery(q)
+	q["status.in_prog"] = true
+	return q
+}
+
+func (d *mongoDriver) getCompletedQuery(q bson.M) bson.M {
```
julianedwards, line 1373: Missing comment.

wait.go
```diff
@@ -99,17 +99,22 @@ func WaitJob(ctx context.Context, j Job, q Queue) bool {
 			return false
 		}
 
+		rj, ok := j.(RetryableJob)
```
julianedwards, line 102: Why not use `amboy.WithRetryableJob` here and below on L134?

queue/driver_mongo.go
```diff
@@ -1260,58 +1250,139 @@ func (d *mongoDriver) scopesInUse(ctx context.Context, scopes []string) bool {
...
+		}
+		if group.ID.InProg {
+			inProg += group.Count
+		}
+		if group.ID.Completed && group.ID.NeedsRetry {
+			retrying += group.Count
+		}
+	}
+
+	// The aggregation cannot also count all the documents in the collection
```
julianedwards, line 1321: Could we use a transaction instead so we can do this completed count atomically?

julianedwards (changes requested):

queue/driver_mongo.go
Kimchelly: I think this could be risky since the Amboy stats job relies on this to be fast and there were some issues with performance with using transactions in Evergreen. I think it's okay for it to be slightly wrong because:
1. The two queries are issued within a short time frame.
2. They're not critical, since they're mostly used for tests (which don't care too deeply about atomicity of stats) and Splunk stats (which don't care that much about # completed jobs).

queue/driver_mongo.go
Kimchelly: Fixed.

queue/driver_mongo.go
Kimchelly: Added.

queue/group_util.go
```diff
@@ -187,6 +188,18 @@ func (c *cacheImpl) Prune(ctx context.Context) error {
 							defer recovery.LogStackTraceAndContinue("panic in queue waiting")
 							defer close(wait)
 
+							// It's possible for a job to be enqueued in between
+							// the queue completeness check and closing the
+							// queue. In this case, job that was just enqueued
```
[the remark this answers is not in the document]
Kimchelly: Fixed.

queue/group_util.go
```diff
@@ -187,6 +188,18 @@ func (c *cacheImpl) Prune(ctx context.Context) error {
...
+							// will not execute unless the queue is restarted.
+							// This log makes it easier to identify if/when this
+							// occurs.
+							grip.Info(message.Fields{
+								"message":  "pruning completed queue",
```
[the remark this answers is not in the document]
Kimchelly: I assume you're talking about [L188](https://github.com/mongodb/amboy/pull/195/files#diff-a11c2e053c1c36554e5dd94c390929edc292e5ae6aba85e0966628263d93f64eR188)? I think that's just a safety mechanism to avoid panicking the entire application. This log just makes it easier to trace when a queue group gets pruned and correlate it with a stranded job, if it ever happens.

I think it could be possible to fix, but I don't think it's that worthwhile, since it would require `Put()` to stop accepting jobs when it's not actively running. This log is mostly for edge case handling.

wait.go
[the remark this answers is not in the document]
Kimchelly: I'm actually going to remove the `RetryableJob` interface and just merge it with the regular `Job` interface in [EVG-14169](https://jira.mongodb.org/browse/EVG-14169), so this should go away.

julianedwards (approved):

PR #198, opened 2021-03-01
EVG-14169: merge RetryableJob into Job interface

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-14169
Patch: https://evergreen.mongodb.com/version/603d5a0f3066154680414970

I was originally hoping to be able to separate the retryable job from the non-retryable job more cleanly. However, it's not really possible to enforce that statically with an interface. It also had a problem where, even if you had a RetryableJob, retryability hinged mostly on the queue's built-in support for it. Instead of relying on a compile-time mechanism to deal with retryable jobs, I'm instead just going to use the `(JobRetryInfo).Retryable` feature flag as the marker for a retryable job.

### DIFF
--- a/queue/driver.go
+++ b/queue/driver.go
@@ -18,10 +18,9 @@ type remoteQueueDriver interface {
 	// Get finds a job by job ID. For retryable jobs, this returns the latest
 	// job attempt.
 	Get(context.Context, string) (amboy.Job, error)
-	// GetAttempt returns a job by job ID and attempt number. Only applicable to
-	// retryable jobs. If used for a non-retryable job, this should return nil
-	// and an error.
-	GetAttempt(ctx context.Context, id string, attempt int) (amboy.RetryableJob, error)
+	// GetAttempt returns a retryable job by job ID and attempt number. If used
+	// to find a non-retryable job, this should return nil job and an error.
+	GetAttempt(ctx context.Context, id string, attempt int) (amboy.Job, error)
 	// Put inserts a new job in the backing storage.
 	Put(context.Context, amboy.Job) error
 	// Save updates an existing job in the backing storage. Implementations may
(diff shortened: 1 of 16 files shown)


queue/driver.go
```diff
@@ -18,10 +18,9 @@ type remoteQueueDriver interface {
 	// Get finds a job by job ID. For retryable jobs, this returns the latest
 	// job attempt.
 	Get(context.Context, string) (amboy.Job, error)
-	// GetAttempt returns a job by job ID and attempt number. Only applicable to
-	// retryable jobs. If used for a non-retryable job, this should return nil
-	// and an error.
-	GetAttempt(ctx context.Context, id string, attempt int) (amboy.RetryableJob, error)
+	// GetAttempt returns a retryable job by job ID and attempt number. If used
+	// to find a non-retryable job, this should return nil job and an error.
```
julianedwards, line 22: Since we are now merging the `RetryableJob` interface with the `Job` interface, would it make more sense for this just to return `nil` and no error when the job is not retryable?

julianedwards, line 22: Since we are now merging the `RetryableJob` interface with the `Job` interface, would it make more sense for this just to return `nil` and no error when the job is not retryable?

julianedwards (changes requested):
Also, would it make sense to merge the `RetryableQueue` with the `Queue` interface? Or would that get too messy?

Kimchelly (commented):
Merging RetryableQueue is undesirable because some queues won't support the atomic guarantee of the retryable queue methods (and wouldn't be worth supporting). Support for retryable jobs is determined by the queue, not the job, which is why I'm throwing out the RetryableJob (all jobs need is for the Retryable flag to be set/unset). The RetryableQueue interface is fine since it indicates optional support for in-memory features.

queue/driver.go
Kimchelly: I don't think it should because squashing legitimate programatic errors (i.e. violating the doc and trying to get a non-retryable job) seems like the wrong move (I also don't like how Evergreen inconsistently returns nil, nil rather than just propagating the error from MongoDB). I also don't want to encourage incorrect usage of the retryable functionality.

julianedwards:
SGTM!

julianedwards (approved):

PR #200, opened 2021-03-02
EVG-14169: export the basic RetryHandler implementation

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-14169

Since the local scope manager is exported, I figured it would be preferable to export the useful implementations of other optional queue components like the retry handler.

Note: I intentionally returned the actual type rather than an interface.

### DIFF
--- a/queue/retry.go
+++ b/queue/retry.go
@@ -105,6 +105,15 @@ func (rh *BasicRetryHandler) Put(ctx context.Context, j amboy.Job) error {
 		return nil
 	}
 
+	grip.Debug(message.Fields{
+		"message":     "put job in retry handler",
+		"job_id":      j.ID(),
+		"job_attempt": j.RetryInfo().CurrentAttempt,
+		"num_pending": len(rh.pending),
+		"queue_id":    rh.queue.ID(),
+		"service":     "amboy.queue.retry",
+	})
+
 	rh.pending[j.ID()] = j
 
 	return nil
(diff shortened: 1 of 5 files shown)


queue/retry.go
```diff
@@ -14,7 +14,10 @@ import (
 	"github.com/pkg/errors"
 )
 
-type basicRetryHandler struct {
+// BasicRetryHandler implements the amboy.RetryHandler interface. It provides a
+// simple component that can be attached to an amboy.RetryableQueue to support
+// automatically retrying jobs.
+type BasicRetryHandler struct {
```
julianedwards, line 20: Does this need to be exported?

julianedwards (changes requested):

queue/retry.go
Kimchelly: Yes. Since `NewBasicRetryHandler` is exported, the type should be exported. Before you say it should return an interface, my preference in Go is to [accept interfaces, return structs](https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a), because it will implicitly satisfy the interface.

julianedwards: I feel so enlightened, thanks for sharing the post.

julianedwards: Also, we should bring this up with the team. We generally return interfaces and maintain impls as unexported.

julianedwards (approved):

queue/retry.go
Kimchelly: 👍 Will bring up in the team meeting or retro.

PR #204, opened 2021-03-04
EVG-14082: Increase logging level for important information

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-14082

I increased the logging level to `info` in `executeJob` when there is no error and an error log's level to `warning` in the mongo go driver.

### DIFF
--- a/pool/helpers.go
+++ b/pool/helpers.go
@@ -79,7 +79,7 @@ func executeJob(ctx context.Context, id string, j amboy.Job, q amboy.Queue) {
 	if err != nil {
 		grip.Error(r)
 	} else {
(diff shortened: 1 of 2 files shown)


Kimchelly (approved):
LGTM, but you can't revendor it because the queue needs new indexes.

julianedwards:
Are the new indexes needed after the retry merge?

Kimchelly:
It's already merged, along with a bunch of other risky changes.

If all you care about is the debug logs, you could just modify the vendored copy of Amboy.

julianedwards:
Yeah I will just do that then, thanks for the head up!

PR #207, opened 2021-03-10
EVG-14245: add method to get all job attempts

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-14245

* Add `GetAllAttempts` to get all attempted executions tied to a single job ID.
* Make retryable queue methods return errors instead of booleans.
* Add error type for job not found in queue and use it for `Save`, `GetAttempt`, and `GetAllAttempts`.
* I decided to move the retrying logic for `CompleteRetrying` out of the remote queue implementation because it seems like that should be handled at a higher level. I only had that because I copied it from `Complete` (which retries because it does not return errors).
* Add tests for RetryableQueue-specific interface.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -19,17 +19,17 @@ func (e *JobNotFoundError) Error() string { return e.msg }
 
 // NewJobNotFoundError creates a new error indicating that a job could not be
 // found in the queue.
-func NewJobNotFoundError(msg string) error { return &JobNotFoundError{msg: msg} }
+func NewJobNotFoundError(msg string) *JobNotFoundError { return &JobNotFoundError{msg: msg} }
 
 // NewJobNotFoundErrorf creates a new error with a formatted message, indicating
 // that a job could not be found in the queue.
-func NewJobNotFoundErrorf(msg string, args ...interface{}) error {
+func NewJobNotFoundErrorf(msg string, args ...interface{}) *JobNotFoundError {
 	return NewJobNotFoundError(fmt.Sprintf(msg, args...))
 }
 
 // MakeJobNotFoundError constructs an error from an existing one, indicating
 // that a job could not be found in the queue.
-func MakeJobNotFoundError(err error) error {
+func MakeJobNotFoundError(err error) *JobNotFoundError {
 	if err == nil {
 		return nil
 	}
--- a/errors.go
+++ b/errors.go
@@ -7,6 +7,43 @@ import (
 	"github.com/pkg/errors"
 )
 
+type JobNotFoundError struct {
+	msg string
+}
+
+func (e *JobNotFoundError) Error() string { return e.msg }
+
+// NewJobNotFoundError creates a new error indicating that a job could not be
+// found in the queue.
+func NewJobNotFoundError(msg string) error { return &JobNotFoundError{msg: msg} }
+
+// NewJobNotFoundErrorf creates a new error with a formatted message, indicating
+// that a job could not be found in the queue.
+func NewJobNotFoundErrorf(msg string, args ...interface{}) error {
+	return NewJobNotFoundError(fmt.Sprintf(msg, args...))
+}
+
+// MakeJobNotFoundError constructs an error from an existing one, indicating
+// that a job could not be found in the queue.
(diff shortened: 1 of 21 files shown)


errors.go
```diff
@@ -7,6 +7,43 @@ import (
 	"github.com/pkg/errors"
 )
 
+type JobNotFoundError struct {
```
julianedwards, line 10: Documentation?

errors.go
```diff
@@ -7,6 +7,43 @@ import (
...
+	msg string
+}
+
+func (e *JobNotFoundError) Error() string { return e.msg }
+
+// NewJobNotFoundError creates a new error indicating that a job could not be
+// found in the queue.
+func NewJobNotFoundError(msg string) error { return &JobNotFoundError{msg: msg} }
```
julianedwards, line 18: Should this return `*JobNotFoundError`?

errors.go
```diff
@@ -7,6 +7,43 @@ import (
...
+// that a job could not be found in the queue.
+func MakeJobNotFoundError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	return NewJobNotFoundError(err.Error())
+}
+
+// IsJobNotFound tests an error it is due tonot being able to find a job in the
```
julianedwards, line 36: I would change this to something like: `IsJobNotFound checks if the error is a job not found error.`

errors.go
```diff
@@ -46,8 +82,7 @@ func MakeDuplicateJobError(err error) error {
 	return NewDuplicateJobError(err.Error())
 }
 
-// IsDuplicateJobError tests an error object to see if it is a
-// duplicate job error.
+// IsDuplicateJobError tests an error to see if it is a duplicate job error.
```
julianedwards, line 85: Can we change `tests` to `checks`?

queue/driver_mongo.go
```diff
@@ -495,32 +495,72 @@ func (d *mongoDriver) Get(ctx context.Context, name string) (amboy.Job, error) {
 	return output, nil
 }
 
-func (d *mongoDriver) GetAttempt(ctx context.Context, name string, attempt int) (amboy.Job, error) {
+// GetAttempt finds a retryable job matching the given job ID and execution
+// attempt. This returns an error if no matching job is found.
+func (d *mongoDriver) GetAttempt(ctx context.Context, id string, attempt int) (amboy.Job, error) {
 	matchIDAndAttempt := bson.M{
-		"retry_info.base_job_id":     name,
+		"retry_info.base_job_id":     id,
```
julianedwards, line 502: We should add an index  `retry_info.base_job_id: 1, retry_info.current_attempt: 1` for this and `GetAllAttempts`.

julianedwards (changes requested):

errors.go
Kimchelly: Added.

errors.go
Kimchelly: It already does.

errors.go
Kimchelly: I tweaked it a bit, but I think it's not that useful if all it does is restate the function's name.

errors.go
Kimchelly: Done.

queue/driver_mongo.go
Kimchelly: Hmm, it might be easier to just use the existing index with `{retry_info.base_job_id: 1, retry_info.current_attempt: -1}`. I modified the logic to make it ascending order in-memory.

errors.go
julianedwards: But the return type is `error`. Aren't we doing the accept interfaces return structs thing?

queue/driver_mongo.go
julianedwards: HA I missed that one! I didn't look up far enough in the file. Sorry.

errors.go
Kimchelly: Oh I see you mean the function signature. In this case, I don't think it matters because the only meaningful way you can use it is as an error. But it's not really an issue to return the struct as-is, so I changed the return value.

julianedwards (approved):

PR #209, opened 2021-03-15
EVG-14076: support retryable jobs in remote management

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-14076

Note: I'm going to probably rework most of the remote management interface for [EVG-14270][EVG-14270], especially how the REST interface and how queue groups are handled, so that it supports more CRUD operations (for ops work) and does less aggregate statistics tracking (not as useful to us). This PR is mostly for the sake of completeness. Nothing is using the remote management interface right now, so it doesn't matter if it's perfect.

* Methods that complete also mark a job as not retrying. This is going to be refactored in [EVG-14270][EVG-14270] to make it easier to deal with retryable jobs and non-retryable jobs.
* Fix a bug where `JobIDsByState` didn't work if `ByGroups` was specified (note: `ByGroups` will likely be removed as part of the rework).
* Change `CompleteJobsByPrefix` to more accurately be `CompleteJobsByPattern` because the MongoDB remote manager's `$regex` is a regexp match, while the queue-backed implementation was a plain prefix match. Also, this makes it more flexible about what job IDs it can match against.
* Include job type info in mongo driver results for `JobInfo()`.
* Fix a bug where the mod time boolean was flipped for the MongoDB remote manager (stale jobs should use `$lte`, not `$gte`).
* Add more tests for remote management interface.

[EVG-14270]: https://jira.mongodb.org/browse/EVG-14270

### DIFF
--- a/cli/queue_management.go
+++ b/cli/queue_management.go
@@ -13,7 +13,7 @@ import (
 const (
 	jobNameFlagName      = "name"
 	jobTypeFlagName      = "type"
-	jobPrefixFlagName    = "job-prefix"
+	jobPatternFlagName   = "job-pattern"
 	statusFilterFlagName = "filter"
 )
 
@@ -28,7 +28,7 @@ func queueManagement(opts *ServiceOptions) cli.Command {
 			managementCompleteJob(opts),
 			managementCompleteJobsByType(opts),
 			managementCompleteJobsByStatus(opts),
-			managementCompleteJobsByPrefix(opts),
+			managementCompleteJobsByPattern(opts),
 		},
 	}
 }
@@ -146,8 +146,8 @@ func managementReportJobIDs(opts *ServiceOptions) cli.Command {
 					if err != nil {
 						return errors.WithStack(err)
 					}
-					for _, j := range report.IDs {
-						t.AddLine(jt, j, report.Group)
+					for _, j := range report.GroupedIDs {
+						t.AddLine(jt, j.Group, j.ID)
 					}
 				}
 				t.Print()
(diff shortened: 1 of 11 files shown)


management/queue.go
```diff
@@ -309,34 +312,54 @@ func (m *queueManager) RecentJobErrors(ctx context.Context, jobType string, wind
...
+		// if it's not a retryable queue, then a retryable job is treated no
+		// differently from a non-retryable job (i.e. it's not retried).
+		if isRetryable {
+			return j, err
+		}
+	}
+
+	j, ok := m.queue.Get(ctx, info.ID)
+	if !ok {
+		return j, errors.New("could not find job")
```
julianedwards, line 337: They should create an `errors.When` for conditional errors.

management/queue.go
```diff
@@ -345,73 +368,85 @@ func (m *queueManager) CompleteJobsByType(ctx context.Context, f StatusFilter, j
 			continue
 		}
 
-		j, ok := m.queue.Get(ctx, info.ID)
-		if !ok {
+		j, err := m.getJob(ctx, info)
+		if err != nil {
+			catcher.Wrapf(err, "job '%s'", info.ID)
```
julianedwards, line 373: Nit: copy error message from L408.

julianedwards (changes requested):

cli/queue_management.go
```diff
@@ -146,8 +146,8 @@ func managementReportJobIDs(opts *ServiceOptions) cli.Command {
 					if err != nil {
 						return errors.WithStack(err)
 					}
-					for _, j := range report.IDs {
-						t.AddLine(jt, j, report.Group)
+					for _, j := range report.GroupedIDs {
+						t.AddLine(jt, j.Group, j.ID)
```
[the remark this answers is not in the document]
Kimchelly: I changed the heading order since results are more likely to be sorted on group.

management/queue.go
Kimchelly: What do you mean? There's no When method in the errors package.

management/queue.go
Kimchelly: Done.

management/queue.go
julianedwards: I know, I was just commenting that it would be great if there were one.

julianedwards (approved):

PR #212, opened 2021-03-22
EVG-14027: add feature flag for retryability

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-14027

* Add feature flag to dynamically disable retryability at runtime.
* Group retryable queue options into common options struct.
* Provide a way to initialize retryable queues with the retry handler in their constructors rather than setting it after construction.
* Tidy up the makefile a bit.

### DIFF
--- a/queue/limited_serializable.go
+++ b/queue/limited_serializable.go
@@ -430,8 +430,9 @@ func (q *limitedSizeSerializableLocal) RetryHandler() amboy.RetryHandler {
 	return q.retryHandler
 }
 
-// SetRetryHandler allows callers to inject a different retry handler
-// implementation if the queue has not yet started.
+// SetRetryHandler allows callers to inject alternative amboy.RetryHandler
+// instances if the queue has not yet started or if the queue is started but
+// does not already have a retry handler.
 func (q *limitedSizeSerializableLocal) SetRetryHandler(rh amboy.RetryHandler) error {
 	q.mu.Lock()
 	defer q.mu.Unlock()
--- a/queue/limited_serializable.go
+++ b/queue/limited_serializable.go
@@ -27,16 +27,15 @@ import (
 // store no more than 2x the specified capacity. For completed jobs, will store
 // no more than the specified capacity.
 type limitedSizeSerializableLocal struct {
-	pending                      chan amboy.Job
-	started                      bool
-	toDelete                     []string
-	capacity                     int
-	storage                      map[string]amboy.Job
-	scopes                       ScopeManager
-	dispatcher                   Dispatcher
-	retryHandler                 amboy.RetryHandler
-	staleRetryingMonitorInterval time.Duration
-	lifetimeCtx                  context.Context
+	pending      chan amboy.Job
+	started      bool
+	toDelete     []string
+	storage      map[string]amboy.Job
+	scopes       ScopeManager
+	dispatcher   Dispatcher
+	retryHandler amboy.RetryHandler
+	opts         LocalLimitedSizeSerializableOptions
+	lifetimeCtx  context.Context
 
 	retryingCount int
 	deletedCount  int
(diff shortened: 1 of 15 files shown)


queue/limited_serializable.go
```diff
@@ -388,13 +422,16 @@ func (q *limitedSizeSerializableLocal) SetRunner(r amboy.Runner) error {
...
+// RetryHandler returns the queue's retry handler.
 func (q *limitedSizeSerializableLocal) RetryHandler() amboy.RetryHandler {
 	q.mu.RLock()
 	defer q.mu.RUnlock()
 
 	return q.retryHandler
 }
 
+// SetRetryHandler allows callers to inject a different retry handler
+// implementation if the queue has not yet started.
```
julianedwards, line 434: `or if the retry handler has not yet been set.`

queue/remote_unordered.go
```diff
@@ -9,25 +9,31 @@ import (
...
-// local worker pool Runner instance of the specified size.
+// configured local worker pool with the specified number of workers.
 func newRemoteUnordered(size int) (remoteQueue, error) {
-	q := &remoteUnordered{
-		remoteBase: newRemoteBase(),
-	}
+	return newRemoteUnorderedWithOptions(remoteOptions{numWorkers: size})
+}
 
+// newRemoteUnorderedWithOptions
```
julianedwards, line 25: Finish documentation.

queue/retry.go
```diff
@@ -9,11 +9,42 @@ import (
...
+func (opts *RetryableQueueOptions) Validate() error {
+	catcher := grip.NewBasicCatcher()
+	catcher.NewWhen(opts.StaleRetryingMonitorInterval < 0, "stale retrying check frequency cannot be negative")
+	if opts.StaleRetryingMonitorInterval == 0 {
+		opts.StaleRetryingMonitorInterval = defaultStaleRetryingMonitorInterval
+	}
+	if opts.Disabled == nil {
+		opts.Disabled = func() bool { return false }
+	}
+	opts.RetryHandler.Disabled = opts.Disabled
```
julianedwards, line 43: What is the point of having a `Disabled` field in `RetryHandlerOptions` if it always gets overwritten? Should this line only occur if `opts.Disabled != nil`?

julianedwards: Also, why does there need to be two `Disabled` fields in the first place? I am a bit confused on how this function would work. Could you give an example?

julianedwards (changes requested):

queue/limited_serializable.go
Kimchelly: It's allowed to overwrite the retry handler even if it's non-nil.

queue/remote_unordered.go
Kimchelly: Done.

queue/retry.go
Kimchelly: 1. Because you can construct a retry handler without this struct and disable it. Disabled will never be nil. This is to handle queue initialization which includes retryable queue options, in which case the retry handler is initialized by the queue constructor. We need the separate setting because monitoring for stale retrying jobs is done at the queue level, not the retry handler level (i.e. `monitorStaleRetryingJobs()`).
2. We need two levels of disabling to deal with retry handler-level disabling and . `opts.Disabled` controls disabling at both the queue level (i.e. monitor stale retrying jobs) and the retry handler level (which is basically just a component of the queue).
3. It has to be a function because it's going to be based on a DB service flag in Evergreen/Cedar that can be toggled at runtime.

queue/limited_serializable.go
julianedwards: But if it has started and the retry handler is non-nil, it can set it.

julianedwards (changes requested):
I think the comment still does not fully reflect the behavior of the function.

queue/limited_serializable.go
Kimchelly: I updated the comment.

julianedwards (approved):

PR #224, opened 2021-04-26
EVG-14519: reduce logging level for Complete errors

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-14519

Reduce log level to warning. I noticed that in evergreen staging, it would log for long-running jobs that failed to complete within the grace period for the app server shutdown, which was noisy when the apps were deployed/restarted. Since jobs  should re-run when they fail to get marked complete, it should be fine to reduce the log level.

### DIFF
--- a/pool/abortable.go
+++ b/pool/abortable.go
@@ -135,7 +135,7 @@ func (p *abortablePool) worker(bctx context.Context) {
 		if err != nil {
 			if job != nil {
 				job.AddError(err)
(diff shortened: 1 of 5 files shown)


julianedwards (approved):

julianedwards:
The only issue with this is that we turn of debug logging for cedar.

Kimchelly:
This is a warning log, so it should still show up in Splunk.

PR #227, opened 2021-05-12
EVG-14609: prevent jobs from being both in progress and completed

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-14609

I've seen cases in staging and prod where jobs can be both in progress and completed at the same time, often for long-running jobs that have their context cancelled during the deploy. This bug seems to have been around for a long while, since I've seen this in jobs going back to at least January, so maybe it's always been a bug but has only been an issue now because more jobs use scopes. I believe this is because `(Job).MarkComplete` marks a job as complete but doesn't clear in progress and `(remoteQueue).Complete` will exit early and will _not_ do anything if the context is cancelled (i.e. when a deploy is shutting down the app server and the queue runner is closed).

Since it's invalid for a job to be both in progress and completed at the same time, I made it more explicit in all places where in progress/completed is set that the two states are mutually exclusive. I also changed the atomic query to make it impossible to update a job back to in progress once it's completed. As a last extra safety measure, I changed the job update helper `doUpdate` to manually intervene and set in progress to false if complete is already true. I'm going to make a Splunk alert for myself to monitor for that log message in case it recurs, since it's a programmer logic error.

### DIFF
--- a/queue/driver_mongo.go
+++ b/queue/driver_mongo.go
@@ -597,30 +597,42 @@ func (d *mongoDriver) Put(ctx context.Context, j amboy.Job) error {
 	return nil
 }
 
-func (d *mongoDriver) getAtomicQuery(jobName string, modCount int) bson.M {
+func (d *mongoDriver) getAtomicQuery(jobName string, stat amboy.JobStatusInfo) bson.M {
 	d.mu.RLock()
 	defer d.mu.RUnlock()
 	owner := d.instanceID
 	timeoutTs := time.Now().Add(-d.LockTimeout())
 
+	// The lock can be acquired if the modification time is unset (i.e. it's
+	// unowned) or older than the lock timeout (i.e. the lock is stale),
+	// regardless of what the other data is.
+	unownedOrStaleLock := bson.M{"status.mod_ts": bson.M{"$lte": timeoutTs}}
+
+	// The lock is actively owned by this in-memory instance of the queue if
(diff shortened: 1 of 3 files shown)


queue/driver_mongo.go
```diff
@@ -803,7 +815,20 @@ func (d *mongoDriver) prepareInterchange(j amboy.Job) (*registry.JobInterchange,
...
+	if ji.Status.InProgress && ji.Status.Completed {
+		err := errors.New("job was found both in progress and complete")
+		grip.Error(message.WrapError(err, (message.Fields{
+			"message":     "programmer error: a job should not be saved as both in progress and complete - manually changing in progress to false",
+			"jira_ticket": "EVG-14605",
+			"job_id":      ji.Name,
+			"service":     "amboy.queue.mdb",
+			"driver_id":   d.ID(),
+		})))
+		ji.Status.InProgress = false
```
julianedwards, line 827: Should this return the error here?

julianedwards (changes requested):

queue/driver_mongo.go
Kimchelly: No. This is a manual correction for a programmatic mistake, and simply informs the programmer to fix their code. It's most likely to happen due to the fact that the dispatcher ping thread and the main thread point to the same in-memory job object.

julianedwards (approved):

PR #230, opened 2021-05-17
EVG-14617: clean up queue resources during close

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-14617

* Close the dispatcher when the queue closes. This avoids two potential issues where a job can be stuck in progress:
    1. The queue might close (when queue groups are pruned or the app restarts), causing all job running threads to finish their jobs and exit, but a job might still dispatch because the dispatcher is still running. This can leave a job stranded in progress because it's been dispatched but there's no job runner to handle it.
    2. The dispatcher might also leave a job stuck in progress forever (until the app restarts) after a runner has already started executing a job. When the queue context is cancelled (e.g. during app shutdown), `Complete` might return early, which prevents the dispatcher from releasing the job.
* Close the driver when the queue closes. Only close the MongoDB client connection if it's owned by the driver (i.e. the driver initialized it and it wasn't passed in with the constructor).
* Cancel all queue threads when the queue closes (e.g. the stale retrying job monitor, the job server thread that populates the next job to be picked up by a job runner).
* Add some more docs.
* Tweak some error and logging messages.

### DIFF
--- a/queue/dispatcher.go
+++ b/queue/dispatcher.go
@@ -202,23 +202,18 @@ func (d *dispatcherImpl) Close(ctx context.Context) error {
 	d.mutex.Lock()
 	defer d.mutex.Unlock()
 
+	catcher := grip.NewBasicCatcher()
 	for jobID := range d.cache {
 		info, ok := d.popInfo(jobID)
 		if !ok {
 			continue
 		}
-		grip.Warning(message.WrapError(d.waitForPing(ctx, info), message.Fields{
-			"message":  "could not wait for job ping to complete",
-			"op":       "release",
-			"job_id":   jobID,
-			"queue_id": d.queue.ID(),
-			"service":  "amboy.queue.dispatcher",
-		}))
+		catcher.Wrapf(d.waitForPing(ctx, info), "waiting for job ping to complete for job '%s'", jobID)
 	}
 
 	d.closed = true
 
-	return nil
+	return catcher.Resolve()
 }
 
 // waitForPing cancels the dispatcher ping for a job and waits for the pinger to
--- a/queue/dispatcher.go
+++ b/queue/dispatcher.go
@@ -26,6 +26,9 @@ type Dispatcher interface {
 	// Complete relinquishes the worker's exclusive ownership of the job. It may
 	// optionally update metadata indicating that the job is finished.
 	Complete(context.Context, amboy.Job)
+	// TODO (EVG-14617): add Close method to stop the dispatcher from further
+	// dispatches. It can be called when the queue closes to stop further
+	// dispatching from occurring and giving up on all actively-dispatched jobs.
 }
 
 type dispatcherImpl struct {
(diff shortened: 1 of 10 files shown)


queue/dispatcher.go
```diff
@@ -187,6 +196,26 @@ func (d *dispatcherImpl) Complete(ctx context.Context, j amboy.Job) {
 	}))
 }
 
+// Close releases all jobs currently locked by the dispatcher and waits for
+// all ping threads to exit.
+func (d *dispatcherImpl) Close(ctx context.Context) error {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+
```
julianedwards, line 204: Should this check if it is already closed an noop if so?

queue/driver_mongo.go
```diff
@@ -90,23 +87,19 @@ func newMongoGroupDriver(name string, opts MongoDBOptions, group string) (remote
 }
 
 // OpenNewMongoGroupDriver constructs and opens a new MongoDB driver instance
-// using the specified session. It is equivalent to calling
-// NewMongoGroupDriver() and calling driver.Open().
-func openNewMongoGroupDriver(ctx context.Context, name string, opts MongoDBOptions, group string, client *mongo.Client) (remoteQueueDriver, error) {
+// using the specified client. The returned driver does not take ownership of
```
julianedwards, line 90: OpenNewMongoGroupDriver -> openNewMongoGroupDriver on the line above.

queue/driver_mongo.go
```diff
@@ -393,10 +396,21 @@ func (d *mongoDriver) reportingIndexes() []mongo.IndexModel {
...
 }
 
-func (d *mongoDriver) Close() {
-	if d.canceler != nil {
-		d.canceler()
+// Close closes the resources for the driver. If it initialized a new client, it
+// is disconnected.
+func (d *mongoDriver) Close(ctx context.Context) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
```
julianedwards, line 403: Should this check if Close was already called?

queue/driver_mongo.go
```diff
@@ -119,43 +112,53 @@ func (d *mongoDriver) ID() string {
...
 		return nil
 	}
+	d.mu.RUnlock()
 
 	client, err := mongo.Connect(ctx, options.Client().ApplyURI(d.opts.URI))
 	if err != nil {
-		return errors.Wrapf(err, "problem opening connection to mongodb at '%s", d.opts.URI)
+		return errors.Wrapf(err, "opening connection to mongodb at '%s", d.opts.URI)
 	}
+	d.ownsClient = true
```
julianedwards, line 129: Should the lock be obtained before updating this?

julianedwards (changes requested):

queue/driver_mongo.go
Kimchelly: Fixed.

queue/driver_mongo.go
Kimchelly: Yeah, you're right. I moved this statement down to the lock in `start` to ensure that both `client` and `ownsClient` are set atomically.

queue/driver_mongo.go
Kimchelly: I added a `closed` check. It has the same issue as the close no-op above, but I don't think it's worth the effort to make `Close` more robust.

julianedwards (approved):

PR #234, opened 2021-06-01
EVG-14644: add option to consider a sample of jobs in Next

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-14644

Currently, `Next` returns an unsorted cursor, so implicitly, it's returning job documents by the order of the index keys. This can lead to contention for the head of the queue, which becomes more contentious the more hosts (i.e. app servers) there are trying to get the next job. I added an option for MongoDB-backed queues to limit the number of jobs considered in each find query, and randomly shuffle the order of the jobs. I'm hoping this will reduce the issue of jobs dispatching too slowly when there are many pending jobs, but I can't be totally certain of the performance impact of this change.

### DIFF
--- a/queue/driver_mongo.go
+++ b/queue/driver_mongo.go
@@ -1133,38 +1133,19 @@ func (d *mongoDriver) getNextQuery() bson.M {
 	d.modifyQueryForGroup(qd)
 
 	timeLimits := bson.M{}
-	if d.opts.CheckWaitUntil && d.opts.CheckDispatchBy {
-		timeLimits["$and"] = []bson.M{
-			d.getWaitUntilQuery(now),
-			d.getDispatchByQuery(now),
+	if d.opts.CheckWaitUntil {
+		timeLimits["time_info.wait_until"] = bson.M{"$lte": now}
+	}
+	if d.opts.CheckDispatchBy {
+		timeLimits["$or"] = []bson.M{
+			{"time_info.dispatch_by": bson.M{"$gt": now}},
+			{"time_info.dispatch_by": time.Time{}},
 		}
-	} else if d.opts.CheckWaitUntil {
-		timeLimits = d.getWaitUntilQuery(now)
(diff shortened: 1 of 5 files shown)


queue/driver_mongo.go
```diff
@@ -1147,9 +1148,43 @@ func (d *mongoDriver) getNextQuery() bson.M {
...
+// same job may be returned multiple times.
+func (d *mongoDriver) getNextSampledPipeline(sampleSize int) []bson.M {
+	match := bson.M{"$match": d.getNextQuery()}
+	// $limit is necessary for performance reasons. $sample scans all input
+	// documents to randomly select documents to return. Therefore, without a
+	// limit on the number of jobs to consider, the cost of $sample will become
+	// more expensive with the number of jobs that it must consider.
+	// Source:
+	// https://docs.mongodb.com/manual/reference/operator/aggregation/sample/#behavior
+	limit := bson.M{"$limit": sampleSize}
```
julianedwards, line 1163: Doesn't the `$limit` with the same number as `$sample` defeat the purpose of `$sample` since it comes after in the pipeline, so it will just "sample" the totality of the documents passed in from the `$limit` stage?

julianedwards: I think it needs to be the first stage in the pipeline, though from the documentation it is not clear if it will respect the following stages (like `$match`):

```
If all the following conditions are met, $sample uses a pseudo-random cursor to select documents:

$sample is the first stage of the pipeline
N is less than 5% of the total documents in the collection
The collection contains more than 100 documents
```

Kimchelly: The `$sample` is used to randomize the document order. `$limit` only limits the number of documents returned by the cursor, but I believe the order is still based on the index key order (so assuming all the queries use the same index to fulfill the query, they would all return the same order).

julianedwards (approved):

PR #236, opened 2021-06-30
EVG-14920: Fix Limited Local Queue Deadlock

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-14920


Kimchelly (approved):

julianedwards:
I fixed a race condition in `Put` where jobs are added to the jobs channel but not to `storage` (since we no longer hold the lock for the entirety of the select statement). I know just add all jobs to `storage` in `Put` but use `pendingStorage` to distinguish from jobs on the queue and jobs waiting in `Put`.

Kimchelly (approved):

PR #240, opened 2021-10-12
EVG-15426: allow scopes to be selectively applied on enqueue

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15426

This modifies the behavior of scopes applied on enqueue. The current behavior is that if you set any scope to be applied on enqueue instead of on dispatch, it applies to all the job's scopes. This provides a mechanism to apply scopes selectively so a job can have some that apply on enqueue and some that apply on dispatch.

* Allow job scopes to be selected to apply on enqueue.
* Rename the apply all scopes on enqueue field for consistency. Once the 90 day TTL has elapsed and there's no jobs with the deprecated field, the deprecated field can be removed. ([EVG-15582](https://jira.mongodb.org/browse/EVG-15582))
* Fix a scope test that had a timing dependency between the dispatcher and the test thread.

### DIFF
--- a/queue/limited_serializable.go
+++ b/queue/limited_serializable.go
@@ -498,7 +498,7 @@ func (q *limitedSizeSerializableLocal) complete(ctx context.Context, j amboy.Job
 		scopesToRelease = j.Scopes()
 	}
 	if err := q.scopes.Release(name, scopesToRelease); err != nil {
-		return errors.Wrapf(err, "releasing scopes '%s'", j.Scopes())
+		return errors.Wrapf(err, "releasing scopes '%s'", scopesToRelease)
 	}
 
 	q.prepareComplete(j)
--- a/registry/interchange.go
+++ b/registry/interchange.go
@@ -18,12 +18,12 @@ type JobInterchange struct {
 	Priority      int                 `json:"priority" bson:"priority" yaml:"priority"`
 	Status        amboy.JobStatusInfo `bson:"status" json:"status" yaml:"status"`
 	Scopes        []string            `bson:"scopes,omitempty" json:"scopes,omitempty" yaml:"scopes,omitempty"`
-	EnqueueScopes []string            `bson:"enqueue_scopes" json:"enqueue_scopes,omitempty" yaml:"enqueue_scopes,omitempty"`
+	EnqueueScopes []string            `bson:"enqueue_scopes,omitempty" json:"enqueue_scopes,omitempty" yaml:"enqueue_scopes,omitempty"`
 	// TODO (EVG-15582): remove this field after all jobs containing this field
 	// have hit their 90 day TTL (2022-02-01) since no jobs will be stored with
 	// this field anymore.
 	EnqueueAllScopesDeprecated bool                   `bson:"apply_scopes_on_enqueue,omitempty" json:"apply_scopes_on_enqueue,omitempty" yaml:"apply_scopes_on_enqueue,omitempty"`
-	EnqueueAllScopes           bool                   `bson:"enqueue_all_scopes" json:"enqueue_all_scopes,omitempty" yaml:"enqueue_all_scopes,omitempty"`
+	EnqueueAllScopes           bool                   `bson:"enqueue_all_scopes,omitempty" json:"enqueue_all_scopes,omitempty" yaml:"enqueue_all_scopes,omitempty"`
 	RetryInfo                  amboy.JobRetryInfo     `bson:"retry_info" json:"retry_info,omitempty" yaml:"retry_info,omitempty"`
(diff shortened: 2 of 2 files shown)


registry/interchange.go
```diff
@@ -11,18 +11,23 @@ import (
...
+	Version       int                 `json:"version" bson:"version" yaml:"version"`
+	Priority      int                 `json:"priority" bson:"priority" yaml:"priority"`
+	Status        amboy.JobStatusInfo `bson:"status" json:"status" yaml:"status"`
+	Scopes        []string            `bson:"scopes,omitempty" json:"scopes,omitempty" yaml:"scopes,omitempty"`
+	EnqueueScopes []string            `bson:"enqueue_scopes" json:"enqueue_scopes,omitempty" yaml:"enqueue_scopes,omitempty"`
+	// TODO (EVG-15582): remove this field after all jobs containing this field
+	// have hit their 90 day TTL (2022-02-01) since no jobs will be stored with
+	// this field anymore.
+	EnqueueAllScopesDeprecated bool                   `bson:"apply_scopes_on_enqueue,omitempty" json:"apply_scopes_on_enqueue,omitempty" yaml:"apply_scopes_on_enqueue,omitempty"`
+	EnqueueAllScopes           bool                   `bson:"enqueue_all_scopes" json:"enqueue_all_scopes,omitempty" yaml:"enqueue_all_scopes,omitempty"`
```
julianedwards, line 26: Should the `bson` tag have an `omitempty`?

registry/interchange.go
```diff
@@ -11,18 +11,23 @@ import (
...
-	Job                  rawJob                 `json:"job" bson:"job" yaml:"job"`
-	Dependency           *DependencyInterchange `json:"dependency,omitempty" bson:"dependency,omitempty" yaml:"dependency,omitempty"`
+	Name          string              `bson:"_id" json:"name" yaml:"name"`
+	Type          string              `json:"type" bson:"type" yaml:"type"`
+	Group         string              `bson:"group,omitempty" json:"group,omitempty" yaml:"group,omitempty"`
+	Version       int                 `json:"version" bson:"version" yaml:"version"`
+	Priority      int                 `json:"priority" bson:"priority" yaml:"priority"`
+	Status        amboy.JobStatusInfo `bson:"status" json:"status" yaml:"status"`
+	Scopes        []string            `bson:"scopes,omitempty" json:"scopes,omitempty" yaml:"scopes,omitempty"`
+	EnqueueScopes []string            `bson:"enqueue_scopes" json:"enqueue_scopes,omitempty" yaml:"enqueue_scopes,omitempty"`
```
julianedwards, line 21: Should the `bson` tag have an `omitempty`?

job/base.go
```diff
@@ -308,23 +310,53 @@ func (b *Base) Scopes() []string {
...
 	b.mutex.RLock()
 	defer b.mutex.RUnlock()
 
-	return b.applyScopesOnEnqueue
+	if b.enqueueAllScopes {
+		return b.RequiredScopes
+	}
+
+	scopes := utility.StringSliceIntersection(b.RequiredScopes, b.enqueueScopes)
+	if len(scopes) == 0 {
```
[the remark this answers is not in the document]
Kimchelly: Yes, this is just an extra safeguard, because nil vs. empty slice matters greatly for checking scopes in the DB. If this can return an empty slice when there's no scopes, the scope `$exists` check will no longer be valid.

queue/limited_serializable.go
```diff
@@ -493,10 +491,14 @@ func (q *limitedSizeSerializableLocal) Complete(ctx context.Context, j amboy.Job
...
-			return errors.Wrapf(err, "releasing scopes '%s'", j.Scopes())
-		}
+	var scopesToRelease []string
+	if j.RetryInfo().ShouldRetry() {
+		scopesToRelease, _ = utility.StringSliceSymmetricDifference(j.Scopes(), j.EnqueueScopes())
+	} else {
+		scopesToRelease = j.Scopes()
+	}
+	if err := q.scopes.Release(name, scopesToRelease); err != nil {
+		return errors.Wrapf(err, "releasing scopes '%s'", j.Scopes())
```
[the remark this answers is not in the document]
Kimchelly: Great catch, you're right. Fixed.

registry/interchange.go
Kimchelly: Yes, it can be omitempty. Changed it.

registry/interchange.go
Kimchelly: Yes, I changed it.

job/base.go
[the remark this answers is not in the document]
Kimchelly: Fair. I think it's easier to understand from a human perspective if the enqueue scopes followed the same expected behavior as the regular scopes field, but it doesn't matter from the DB's perspective.

PR #251, opened 2021-12-02
EVG-14869: provide default dependency manager for jobs

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-14869

### Description
* Make the dependency manager optional. It's only used in a few queue implementations (none of which is used). It should no longer be necessary to always include `job.SetDependency(dependency.NewAlways())` in the job constructor.
* Default the dependency manager to use `dependency.NewAlways()`. This is necessary because the job interchange and some queue implementations already assume that the job dependency manager is never nil.

### Testing
* Remove `job.SetDependency(dependency.NewAlways())` from all tests and confirm that everything works as it did previously.


job/base.go
```diff
@@ -169,20 +169,30 @@ func (b *Base) Type() amboy.JobType {
...
 
 // SetDependency allows you to inject a different Job dependency
-// object, and is a component of the Job interface.
+// object, and is a component of the Job interface. If the given dependency
+// manager is nil, it will default to the always manager.
 func (b *Base) SetDependency(d dependency.Manager) {
 	b.mutex.Lock()
 	defer b.mutex.Unlock()
 
+	if d == nil {
```
julianedwards, line 192: This seems unnecessary since you made the change to lazily set nil dependency manager to the always manager in `Dependency()` above.

julianedwards (changes requested):

job/base.go
Kimchelly: Changed it.

julianedwards (approved):

PR #258, opened 2022-01-06
EVG-15956: clarify logged timing statistics when job completes

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15956

Using the word "pending" in this case is deceptive because pending in an Amboy job refers to the state when the job is waiting in the queue to be dispatched to a worker. Here, it's being used to refer to the job's turnaround time (i.e. timespan between job submission and completion of its work). To keep the nomenclature consistent, I renamed the log field.

Once this is merged and Amboy is bumped in Evergreen/Cedar, I'll fix the Splunk Amboy dashboards that currently use "pending_secs" to populate the charts.


hhoke (approved):
LGTM

PR #259, opened 2022-01-26
EVG-15931: support per-queue options in queue group

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15931

* Heavily refactor and reduce complexity of options to reduce the amount of confusion and duplication in queue group, queue, and driver. As far as I can tell, this is logically equivalent to the existing logic, but some of the duplicated options are consolidated. When this is revendored, I'll rename options fields accordingly in Evergreen/Cedar.
    * Move some options structs into more natural file locations (e.g. `MongoDBQueueOptions`).
    * Consolidate how remote queue groups create queues vs. how standalone remote queues are created so they all use `MongoDBQueueOptions.BuildQueue`.
    * Consolidate various driver constructors.
    * Push duplicated fields in queue groups into the queue level.
    * Rename the vague driver "name" to "collection", since that's what it actually is.
* Add extra safety checks to queue group implementations to ensure that they set the proper group settings for the driver.
* Allow `(amboy.QueueGroup).Get` to accept queue options. Each queue group constructor initializes with a default set of queue options which are applied to the queues if the queue-specific options are not provided. The queue-level options take precedence over the default options.
* Add more validation for input options.
* Add more documentation for exported things.


queue/remote.go
```diff
@@ -3,58 +3,77 @@ package queue
...
+// persistence layer to support distributed systems of workers.
+type MongoDBQueueOptions struct {
+	// DB represents options for the MongoDB driver.
+	DB *MongoDBOptions
+	// NumWorkers is the default number of workers the queue should use. This
+	// has lower precedence than WorkerPoolSize.
+	NumWorkers *int
+	// WorkerPoolSize returns the number of workers the queue should use. If
+	// set, this takes precedence over NumWorkers.
+	WorkerPoolSize func(string) int
```
hhoke, line 23: What is `func(string)` doing here? I don't understand.

queue/group_remote_mongo_single.go
```diff
@@ -10,40 +10,46 @@ import (
...
 
-// NewMongoDBSingleQueueGroup constructs a new remote queue group. If ttl is 0, the queues will not be
-// TTLed except when the client explicitly calls Prune.
-func NewMongoDBSingleQueueGroup(ctx context.Context, opts MongoDBQueueGroupOptions, client *mongo.Client, mdbopts MongoDBOptions) (amboy.QueueGroup, error) {
+// NewMongoDBSingleQueueGroup constructs a new remote queue group. If the TTL is
+// 0, the queues will not be TTLed except when the client explicitly calls
+// Prune.
+//
+// The MongoDB single remote queue group multiplexes all queues into a single
+// collection. Group name as a means to namespace each queue within the
```
julianedwards, line 29: `Group name is used as a means...`

queue/group_local.go
```diff
@@ -70,16 +117,26 @@ func (g *localQueueGroup) Queues(_ context.Context) []string {
...
 	if q != nil {
 		return q, nil
 	}
 
-	queue, err := g.opts.Constructor(ctx)
+	localQueueOpts, err := getLocalQueueOptions(opts...)
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid queue options")
+	}
+	queueOpts := mergeLocalQueueOptions(append([]LocalQueueOptions{g.opts.Queue}, localQueueOpts...)...)
```
julianedwards, line 135: Should the passed in opts take precedence here?

queue/group_remote_mongo.go
```diff
@@ -231,33 +173,34 @@ func (g *remoteMongoQueueGroup) Queues(ctx context.Context) []string {
...
-		}
-	} else {
-		d, err = newMongoDriver(coll, g.dbOpts)
-		if err != nil {
-			return nil, errors.Wrap(err, "creating and opening driver")
-		}
-	}
-	if err := q.SetDriver(d); err != nil {
-		return nil, errors.Wrap(err, "setting driver")
+	queueOpts := mergeMongoDBQueueOptions(append([]MongoDBQueueOptions{g.opts.Queue}, mdbOpts...)...)
```
julianedwards, line 188: Should passed in opts take precedence?

julianedwards (changes requested):

queue/group_remote_mongo.go
```diff
@@ -201,7 +143,7 @@ func (g *remoteMongoQueueGroup) startQueues(ctx context.Context) error {
 	g.mu.Lock()
 	defer g.mu.Unlock()
 
-	colls, err := g.getExistingCollections(ctx, g.client, g.dbOpts.DB, g.opts.Prefix)
+	colls, err := g.getExistingCollections(ctx, g.opts.Queue.DB.Client, g.opts.Queue.DB.DB, g.opts.Queue.DB.GroupName)
```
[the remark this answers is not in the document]
Kimchelly: Nice catch, you're absolutely right. This is a bit tricky since this implementation previously used the prefix (which I removed) as both a prefix for each collection's name and for the individual jobs within each collection. I removed the prefix because the individual jobs in the collection don't need the prefix, seeing how they're already isolated by being in separate collections. However, I think we might still need the collection-level prefix, since we can't necessarily assume that all the collections in the DB are just for this queue group.

I re-added the collection-level prefix parameter.

queue/group_remote_mongo.go
Kimchelly: Same thing as above - later options take precedence over earlier ones.

queue/group_remote_mongo.go
```diff
@@ -231,33 +173,34 @@ func (g *remoteMongoQueueGroup) Queues(ctx context.Context) []string {
...
-		return nil, errors.Wrap(err, "setting driver")
+	queueOpts := mergeMongoDBQueueOptions(append([]MongoDBQueueOptions{g.opts.Queue}, mdbOpts...)...)
+
+	// The collection name has to be set to ensure the queue uses its
+	// collection-level namespace.
+	queueOpts.DB.Collection = coll
+	// These settings must apply to all queues in the queue group because if the
+	// queue-specific options differ from the defaults, it will affect the queue
+	// group's ability to manage jobs properly.
+	queueOpts.DB.UseGroups = g.opts.Queue.DB.UseGroups
```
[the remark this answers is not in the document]
Kimchelly: Yeah, this is saying that they have to remain unset (or whatever is the queue group-level setting). Even though they're configurable as queue-level options, they have to remain consistent between the queue group and queue.

queue/group_remote_mongo_single.go
Kimchelly: Great catch! I removed this `GroupName` comment since it's not actually pertinent to the constructor's caller. I might eventually make `UseGroups` and `GroupName` unexported since they should really only be set by the queue groups and not by end users.

Kimchelly (commented):
I also added a bit of extra validation for queue groups to avoid panics in case of nil/unset options that are required.

queue/group_local.go
julianedwards: Haha how embarrassing, I read `mergeLocalQueueOptions` incorrectly. I thought it was only setting it if it wasn't already set.

julianedwards (approved):
Nice work! LGTM!

PR #262, opened 2022-02-07
EVG-16210: add per-queue initialization options for the queue group

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16210

When the queue group is first started, it initializes queues that already exist and are active (i.e. have incomplete jobs) in the cache based on the DB state. However, the cached queues don't have the ability to set their queue options after the queue is already in the cache since it's considered active and running. This provides a feature to allow users to specify the initial options for individual queues during queue group initialization to ensure that they get the desired options.

* Add per-queue options that can be set during queue group initialization.
* Rename the default queue options to avoid ambiguity.

### DIFF
--- a/queue/group_remote_mongo.go
+++ b/queue/group_remote_mongo.go
@@ -57,16 +57,30 @@ type MongoDBQueueGroupOptions struct {
 	TTL time.Duration
 }
 
-func (opts MongoDBQueueGroupOptions) validate() error {
+func (o MongoDBQueueGroupOptions) validate() error {
 	catcher := grip.NewBasicCatcher()
-	catcher.NewWhen(opts.TTL < 0, "TTL must be greater than or equal to 0")
-	catcher.NewWhen(opts.TTL > 0 && opts.TTL < time.Second, "TTL cannot be less than 1 second, unless it is 0")
-	catcher.NewWhen(opts.PruneFrequency < 0, "prune frequency must be greater than or equal to 0")
-	catcher.NewWhen(opts.PruneFrequency > 0 && opts.TTL < time.Second, "prune frequency cannot be less than 1 second, unless it is 0")
-	catcher.NewWhen((opts.TTL == 0 && opts.PruneFrequency != 0) || (opts.TTL != 0 && opts.PruneFrequency == 0), "ttl and prune frequency must both be 0 or both be not 0")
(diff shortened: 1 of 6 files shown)


queue/group_remote_mongo.go
```diff
@@ -196,22 +200,30 @@ func (g *remoteMongoQueueGroup) Queues(ctx context.Context) []string {
...
 	if err != nil {
 		return nil, errors.Wrap(err, "getting queue options")
 	}
 
-	queueOpts := mergeMongoDBQueueOptions(append([]MongoDBQueueOptions{g.opts.Queue}, mdbOpts...)...)
+	precedenceOrderedOpts := []MongoDBQueueOptions{g.opts.DefaultQueue}
+	if perQueueOpts, ok := g.opts.PerQueue[id]; ok {
+		precedenceOrderedOpts = append(precedenceOrderedOpts, perQueueOpts)
+	}
+	precedenceOrderedOpts = append(precedenceOrderedOpts, mdbOpts...)
```
julianedwards, line 216: So just to be clear: the order of precedence is (1) mdb options, (2) per que opts, and (3) default?

queue/group_remote_mongo_single.go
```diff
@@ -188,14 +184,22 @@ func (g *remoteMongoQueueGroupSingle) Get(ctx context.Context, id string, opts .
 		if err != nil {
 			return nil, errors.Wrap(err, "getting queue options")
 		}
-		queueOpts = mergeMongoDBQueueOptions(append([]MongoDBQueueOptions{g.opts.Queue}, mdbOpts...)...)
+
+		precedenceOrderedOpts := []MongoDBQueueOptions{g.opts.DefaultQueue}
```
julianedwards, line 188: This block of code is reused, can we modulate it into a helper function?

julianedwards (changes requested):

queue/group_remote_mongo.go
Kimchelly: Yes.

queue/group_remote_mongo_single.go
Kimchelly: Done.

julianedwards (approved):

PR #287, opened 2022-08-01
EVG-17370: support named queue settings matching a regexp

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17370

We already have options that can be applied to specific named queues in the queue group. This adds the ability to apply queue settings to named queues that match a regexp instead of an exact matching name.

### DIFF
--- a/queue/group_remote_mongo.go
+++ b/queue/group_remote_mongo.go
@@ -3,6 +3,7 @@ package queue
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"runtime"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


queue/group_remote_mongo.go
```diff
@@ -34,8 +35,13 @@ type MongoDBQueueGroupOptions struct {
...
 	DefaultQueue MongoDBQueueOptions
 
+	// RegexpQueue represents options for queues by IDs that match a regular
+	// expression. These take precedence over the DefaultQueue options but have
+	// lower precedence than the PerQueue options.
+	RegexpQueue []RegexpMongoDBQueueOptions
+
 	// PerQueue represent options for specific queues by ID. These take
-	// precedence over the DefaultQueue options.
+	// precedence over the DefaultQueue and PerQueueRegexp options.
```
julianedwards, line 44: `QueueRegexp`

queue/group_remote_mongo.go
```diff
@@ -57,6 +63,15 @@ type MongoDBQueueGroupOptions struct {
 	TTL time.Duration
 }
 
+// RegexpMongoDBQueueOptions represents a mapping from a regular expression to
+// match named queues in a queue group to options for those queues.
+type RegexpMongoDBQueueOptions struct {
+	// Regexp is the regular expression to match against the queue ID.
+	Regexp regexp.Regexp
```
julianedwards, line 70: Would it make more sense to make this a pointer?

julianedwards (changes requested):

queue/group_remote_mongo.go
Kimchelly: Fixed, thanks for the catch!

queue/group_remote_mongo.go
Kimchelly: I guess it could be made a pointer, but it's not an optional field, so I was inclined to not let it be potentially nil.

julianedwards (approved):

PR #299, opened 2022-10-25
EVG-18129: add option to explicitly specify next job hint

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18129

Make the index pattern hint for the next job query configurable by the user when setting up the MongoDB driver. We could use this in the future as well if we want to explicitly set indexes for other queries. By default, it doesn't hint any index for the query, so applications can pick their desired index according to their needs.

### DIFF
--- a/queue/driver_mongo.go
+++ b/queue/driver_mongo.go
@@ -92,7 +92,7 @@ type MongoDBOptions struct {
 
 // PreferredIndexOptions provide options to explicitly set the index for use in
 // specific scenarios. If an index is not explicitly given, the index will be
-// picked auomatically.
+// picked automatically.
 type PreferredIndexOptions struct {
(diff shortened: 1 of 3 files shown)


queue/driver_mongo.go
```diff
@@ -87,6 +90,17 @@ type MongoDBOptions struct {
 	SampleSize int
 }
 
+// PreferredIndexOptions provide options to explicitly set the index for use in
+// specific scenarios. If an index is not explicitly given, the index will be
+// picked auomatically.
```
ybrill, line 95: automatically

ybrill (approved):
This is much cleaner than the hotfix 😄

queue/driver_mongo.go
Kimchelly: Fixed, thanks!

PR #300, opened 2022-10-27
EVG-18128: check wait until in-memory

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18128

`CheckWaitUntil` will omit jobs that haven't yet reached their `WaitUntil` requirement from the next job iterator results. However, if `CheckWaitUntil` is disabled, jobs that haven't hit their `WaitUntil` time will appear in the iterator results, so the queue should fall back to checking if `WaitUntil` is satisfied in-memory.

### DIFF
--- a/queue/driver_test.go
+++ b/queue/driver_test.go
@@ -945,11 +945,11 @@ func TestDriverNextJob(t *testing.T) {
 			mq := makeMockRemoteQueue(t, mdbOpts)
 
 			for testName, testCase := range map[string]func(ctx context.Context, t *testing.T, driver *mongoDriver){
-				"NextReturnsNoJobIfNoneExists": func(ctx context.Context, t *testing.T, driver *mongoDriver) {
+				"ReturnsNoJobIfNoneExists": func(ctx context.Context, t *testing.T, driver *mongoDriver) {
 					j := driver.Next(ctx)
 					assert.Nil(t, j)
 				},
-				"NextResolvesJobContentionAndDispatchesToSingleWorker": func(ctx context.Context, t *testing.T, driver *mongoDriver) {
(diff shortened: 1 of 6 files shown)


queue/driver_test.go
```diff
@@ -1005,7 +1005,7 @@ func TestDriverNextJob(t *testing.T) {
 
 					checkDispatched(t, next.Status())
 				},
-				"NextIgnoresJobThatHasNotHitWaitUntilRequirementYet": func(ctx context.Context, t *testing.T, driver *mongoDriver) {
+				"IgnoresJobThatHasNotHitWaitUntilRequirementYet": func(ctx context.Context, t *testing.T, driver *mongoDriver) {
```
ybrill, line 1008: could it make sense to test this with a driver with `CheckWaitUntil == false`?

ybrill (approved):
I don't really understand why the driver supports ignoring `WaitUntil` (and `DispatchBy`). It looks like maybe Sam intended it as like a feature flag back in https://github.com/mongodb/amboy/pull/22? Could we remove these options? It seems like it wouldn't affect any of _our_ repos.

Kimchelly (commented):
I think it optionally ignores them for backward compatibility + application-specific performance reasons because the indexes necessary to fulfill them may not exist in all applications (and may not be wanted). Having an in-memory check is supposed to make the feature work regardless of whether the application has the necessary indexes to filter them in the query.

I also remember that enabling `CheckDispatchBy` is particularly problematic for performance because it can create a large graveyard of pending jobs that slowly degrades the efficiency of the pending job filter (unless there's a mechanism like what the next job loop currently does to delete pending jobs exceeding their dispatch by limit).

queue/driver_test.go
Kimchelly: Sure, I added it.

PR #307, opened 2022-12-01
EVG-18304: allow the local limited size queue to accept jobs before processing starts

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18304

The local limited size queue will currently return an error if you try to enqueue jobs before the workers have started processing the queue. This isn't really necessary (and deviates from how all the other implementations behave), and it should be possible to pre-populate the queue before starting work on it, so I removed this limitation. This feature is useful for testing.


opsnlops:
I don't know enough about this code to feel comfortable reviewing it, but I like this idea!

julianedwards (approved):

PR #308, opened 2022-12-05
EVG-18240 break up Next query

ybrill:
[EVG-18240](https://jira.mongodb.org/browse/EVG-18240)

Break the query in two: the stale in-progress part first and, if anything's left of the sample size, the pending part.

It seemed the non-sample case could also be an aggregation so this combines them. It skips the check for exclusivity of sample size and priority since it's an invalid configuration anyways (did I get that right?).

### DIFF
--- a/queue/driver_mongo.go
+++ b/queue/driver_mongo.go
@@ -1288,7 +1288,11 @@ func (d *mongoDriver) dispatchStaleAndPending(ctx context.Context, startAt time.
 		return job, info, nil
 	}
 
-	job, dispatchInfo, err = d.tryDispatchJob(ctx, d.getNextPendingQuery(now), d.opts.SampleSize-dispatchInfo.total, startAt)
+	pendingSampleSize := 0
+	if d.opts.SampleSize > 0 {
+		pendingSampleSize = d.opts.SampleSize - dispatchInfo.total
+	}
+	job, dispatchInfo, err = d.tryDispatchJob(ctx, d.getNextPendingQuery(now), pendingSampleSize, startAt)
 	info.misses += dispatchInfo.misses
 	info.skips += dispatchInfo.skips
 	if err != nil {
@@ -1300,7 +1304,7 @@ func (d *mongoDriver) dispatchStaleAndPending(ctx context.Context, startAt time.
 	return job, info, nil
 }
 
-// getNextQuery returns the query for the next available jobs.
+// getNextStaleInProgressQuery returns the query for the next available stale in-progress jobs.
 func (d *mongoDriver) getNextStaleInProgressQuery(now time.Time) bson.M {
 	staleBefore := now.Add(-d.LockTimeout())
 	qd := d.getInProgQuery(bson.M{"status.mod_ts": bson.M{"$lte": staleBefore}})
@@ -1310,6 +1314,7 @@ func (d *mongoDriver) getNextStaleInProgressQuery(now time.Time) bson.M {
 	return qd
 }
 
+// getNextPendingQuery returns the query for the next available pending jobs.
 func (d *mongoDriver) getNextPendingQuery(now time.Time) bson.M {
 	qd := d.getPendingQuery(bson.M{})
 	d.modifyQueryForGroup(qd)
@@ -1363,7 +1368,7 @@ func (d *mongoDriver) tryDispatchJob(ctx context.Context, query bson.M, sampleSi
 	return job, dispatchInfo, nil
 }
 
-// getNextCursor returns a cursor for the next available jobs matching nextQuery.
+// getNextCursor returns a cursor for the jobs matching nextQuery.
 // If sampleSize is greater than zero it's added as a limit and the jobs are shuffled.
 func (d *mongoDriver) getNextCursor(ctx context.Context, nextQuery bson.M, sampleSize int) (*mongo.Cursor, error) {
 	pipeline := []bson.M{{"$match": nextQuery}}
--- a/queue/driver_mongo.go
+++ b/queue/driver_mongo.go
@@ -1308,8 +1308,7 @@ func (d *mongoDriver) getNextStaleInProgressQuery(now time.Time) bson.M {
 // getNextPendingQuery returns the query for the next available pending jobs.
 func (d *mongoDriver) getNextPendingQuery(now time.Time) bson.M {
 	qd := d.getPendingQuery(bson.M{})
-	d.modifyQueryForGroup(qd)
-	qd = d.modifyQueryForTimeLimits(qd, now)
+	d.modifyQueryForGroup(d.modifyQueryForTimeLimits(qd, now))
 
 	return qd
 }
--- a/queue/driver_mongo.go
+++ b/queue/driver_mongo.go
@@ -1191,108 +1191,6 @@ func (d *mongoDriver) JobInfo(ctx context.Context) <-chan amboy.JobInfo {
 	return infos
 }
 
-// getNextQuery returns the query for the next available jobs.
-func (d *mongoDriver) getNextQuery() bson.M {
-	now := time.Now()
-	staleBefore := now.Add(-d.LockTimeout())
-	qd := bson.M{
-		"$or": []bson.M{
-			d.getPendingQuery(bson.M{}),
-			d.getInProgQuery(
-				bson.M{"status.mod_ts": bson.M{"$lte": staleBefore}},
-			),
-		},
-	}
-
-	d.modifyQueryForGroup(qd)
-
-	var timeLimits []bson.M
-	if d.opts.CheckWaitUntil {
-		checkWaitUntil := bson.M{"$or": []bson.M{
-			{"time_info.wait_until": bson.M{"$lte": now}},
-			{"time_info.wait_until": bson.M{"$exists": false}},
-		}}
-		timeLimits = append(timeLimits, checkWaitUntil)
-	}
-
-	if d.opts.CheckDispatchBy {
-		checkDispatchBy := bson.M{"$or": []bson.M{
-			{"time_info.dispatch_by": bson.M{"$gt": now}},
-			{"time_info.dispatch_by": bson.M{"$exists": false}},
-		}}
-		timeLimits = append(timeLimits, checkDispatchBy)
-	}
-
-	if len(timeLimits) > 0 {
-		qd = bson.M{"$and": append(timeLimits, qd)}
-	}
-
-	return qd
-}
-
-// getNextSampledPipeline returns an aggregation pipeline to query for the next
-// available jobs, with a maximum limit (sampleSize) on the number of jobs
-// considered and returned. Jobs are returned in shuffled random order, and the
-// same job may be returned multiple times.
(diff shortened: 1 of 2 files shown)


queue/driver_mongo.go
```diff
@@ -1191,108 +1191,6 @@ func (d *mongoDriver) JobInfo(ctx context.Context) <-chan amboy.JobInfo {
...
-		qd = bson.M{"$and": append(timeLimits, qd)}
-	}
-
-	return qd
-}
-
-// getNextSampledPipeline returns an aggregation pipeline to query for the next
-// available jobs, with a maximum limit (sampleSize) on the number of jobs
-// considered and returned. Jobs are returned in shuffled random order, and the
-// same job may be returned multiple times.
```
ybrill, line 1236: Calling out: this removed the 
> and the same job may be returned multiple times

bit since I didn't understand how that would be.

queue/driver_mongo.go
```diff
@@ -1400,14 +1268,146 @@ type dispatchAttemptInfo struct {
 	// misses is the number of times the dispatcher attempted to dispatch a job
 	// but failed.
 	misses int
+	// total is the total number of documents the cursor returned.
+	total int
+}
+
+// dispatchStaleAndPending attempts to dispatch a job. Stale in progress jobs are given precedence over pending jobs.
+func (d *mongoDriver) dispatchStaleAndPending(ctx context.Context, startAt time.Time) (amboy.Job, dispatchAttemptInfo, error) {
```
Kimchelly, line 1276: `dispatchNextJob`?

queue/driver_mongo.go
```diff
@@ -1400,14 +1268,146 @@ type dispatchAttemptInfo struct {
...
+		q = bson.M{"$and": append(timeLimits, q)}
+	}
+
+	return q
+}
+
+// tryDispatchJob attempts to dispatch a job that matches the query. If sampleSize is greater than
+// zero only that number of jobs are considered and the jobs are randomly shuffled.
+// If dispatching succeeds, the successfully dispatched job is returned.
+func (d *mongoDriver) tryDispatchJob(ctx context.Context, query bson.M, sampleSize int, startAt time.Time) (amboy.Job, dispatchAttemptInfo, error) {
```
Kimchelly, line 1356: `tryDispatchWithQuery`?

queue/driver_mongo.go
```diff
@@ -1400,14 +1268,146 @@ type dispatchAttemptInfo struct {
...
+	var info dispatchAttemptInfo
+
+	job, dispatchInfo, err := d.tryDispatchJob(ctx, d.getNextStaleInProgressQuery(now), d.opts.SampleSize, startAt)
+	info.misses += dispatchInfo.misses
+	info.skips += dispatchInfo.skips
+	if err != nil {
+		return nil, info, errors.Wrap(err, "dispatching stale in progress job")
+	}
+
+	if job != nil || d.opts.SampleSize > 0 && dispatchInfo.total >= d.opts.SampleSize {
```
Kimchelly, line 1287: It seems like it would be preferable to continue trying to dispatch pending jobs even if it already iterated `SampleSize` # of in progress jobs to avoid overly blocking the pending queue on the stale in progress jobs.

queue/driver_mongo.go
```diff
@@ -1400,14 +1268,146 @@ type dispatchAttemptInfo struct {
...
+}
+
+// dispatchStaleAndPending attempts to dispatch a job. Stale in progress jobs are given precedence over pending jobs.
+func (d *mongoDriver) dispatchStaleAndPending(ctx context.Context, startAt time.Time) (amboy.Job, dispatchAttemptInfo, error) {
+	now := time.Now()
+	var info dispatchAttemptInfo
+
+	job, dispatchInfo, err := d.tryDispatchJob(ctx, d.getNextStaleInProgressQuery(now), d.opts.SampleSize, startAt)
+	info.misses += dispatchInfo.misses
+	info.skips += dispatchInfo.skips
```
Kimchelly, line 1282: Is this also supposed to add `total` to the count?

queue/driver_mongo.go
```diff
@@ -1400,14 +1268,146 @@ type dispatchAttemptInfo struct {
 	// misses is the number of times the dispatcher attempted to dispatch a job
 	// but failed.
 	misses int
+	// total is the total number of documents the cursor returned.
```
Kimchelly, line 1271: total is the total number of jobs that were considered for dispatch.

queue/driver_mongo.go
```diff
@@ -1400,14 +1268,146 @@ type dispatchAttemptInfo struct {
...
+	}
+
+	return job, info, nil
+}
+
+// getNextStaleInProgressQuery returns the query for the next available stale in-progress jobs.
+func (d *mongoDriver) getNextStaleInProgressQuery(now time.Time) bson.M {
+	staleBefore := now.Add(-d.LockTimeout())
+	qd := d.getInProgQuery(bson.M{"status.mod_ts": bson.M{"$lte": staleBefore}})
+	d.modifyQueryForGroup(qd)
```
Kimchelly, line 1311: Just a suggestion that you could turn this into a nicer-looking builder if `modifyQueryForGroup` returns the `bson.M` query to do `d.modifyQueryForGroup(d.modifyQueryForTimeLimits(qd, now))`.

queue/driver_mongo.go
```diff
@@ -1333,54 +1231,24 @@ func (d *mongoDriver) Next(ctx context.Context) amboy.Job {
...
-					"duration_secs": time.Since(startAt).Seconds(),
-				}))
-				return nil
-			}
-
-			job, dispatchInfo := d.tryDispatchJob(ctx, iter, startAt)
-			dispatchMisses += dispatchInfo.misses
-			dispatchSkips += dispatchInfo.skips
-			if err = iter.Err(); err != nil {
 				if job != nil {
```
Kimchelly, line 1236: Seems like this could never happen since the function returns either `err` or `job`, but not both.

queue/driver_mongo.go
```diff
@@ -1333,54 +1231,24 @@ func (d *mongoDriver) Next(ctx context.Context) amboy.Job {
...
-					"driver_id": d.instanceID,
-					"service":   "amboy.queue.mdb",
-					"operation": "retrieving next job",
-					"is_group":  d.opts.UseGroups,
-					"group":     d.opts.GroupName,
-				}))
-				return nil
-			}
+			dispatchMisses += dispatchInfo.misses
+			dispatchSkips += dispatchInfo.skips
```
Kimchelly, line 1251: Should this include the total # jobs?

queue/driver_mongo.go
```diff
@@ -1400,14 +1268,146 @@ type dispatchAttemptInfo struct {
...
+	if err != nil {
+		return nil, info, errors.Wrap(err, "dispatching stale in progress job")
+	}
+
+	if job != nil || d.opts.SampleSize > 0 && dispatchInfo.total >= d.opts.SampleSize {
+		return job, info, nil
+	}
+
+	pendingSampleSize := 0
+	if d.opts.SampleSize > 0 {
```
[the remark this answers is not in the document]
Kimchelly: I also think it's okay to not reduce the sample size for the follow-up query. The main purpose of the field is to cap the number of jobs checked per query, not cap the total number of jobs checked per `Next` round. We can just change the doc comment on `SampleSize` to indicate that it's the number of jobs per set of jobs checked rather than total.

queue/driver_mongo.go
```diff
@@ -1400,14 +1268,146 @@ type dispatchAttemptInfo struct {
...
+		pendingSampleSize = d.opts.SampleSize - dispatchInfo.total
+	}
+	job, dispatchInfo, err = d.tryDispatchJob(ctx, d.getNextPendingQuery(now), pendingSampleSize, startAt)
+	info.misses += dispatchInfo.misses
+	info.skips += dispatchInfo.skips
+	if err != nil {
+		if err != nil {
```
Kimchelly, line 1299: Remove extra nil err check?

queue/driver_mongo.go
```diff
@@ -1400,14 +1268,146 @@ type dispatchAttemptInfo struct {
...
+// If dispatching succeeds, the successfully dispatched job is returned.
+func (d *mongoDriver) tryDispatchJob(ctx context.Context, query bson.M, sampleSize int, startAt time.Time) (amboy.Job, dispatchAttemptInfo, error) {
+	iter, err := d.getNextCursor(ctx, query, sampleSize)
+	if err != nil {
+		return nil, dispatchAttemptInfo{}, errors.Wrap(err, "getting next job iterator")
+	}
+
+	job, dispatchInfo := d.tryDispatchFromCursor(ctx, iter, startAt)
+
+	if err = iter.Err(); err != nil {
```
Kimchelly, line 1364: If this errors, we have to still ensure we release the job from the dispatcher if it's non-nil.

queue/driver_mongo.go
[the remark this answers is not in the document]
ybrill: done.
That made it a lot easier!

queue/driver_mongo.go
ybrill: done

queue/driver_mongo.go
ybrill: done

queue/driver_mongo.go
ybrill: done

queue/driver_mongo.go
ybrill: total went away

queue/driver_mongo.go
ybrill: total went away

queue/driver_mongo.go
ybrill: `dispatchNextJob` returns the job now

queue/driver_mongo.go
ybrill: total went away

queue/driver_mongo.go
ybrill: oops.
removed

queue/driver_mongo.go
ybrill: returned the job

queue/driver_mongo.go
```diff
@@ -1402,12 +1270,133 @@ type dispatchAttemptInfo struct {
...
+	qd := d.getInProgQuery(bson.M{"status.mod_ts": bson.M{"$lte": staleBefore}})
+	d.modifyQueryForGroup(d.modifyQueryForTimeLimits(qd, now))
+
+	return qd
+}
+
+// getNextPendingQuery returns the query for the next available pending jobs.
+func (d *mongoDriver) getNextPendingQuery(now time.Time) bson.M {
+	qd := d.getPendingQuery(bson.M{})
+	d.modifyQueryForGroup(qd)
```
Kimchelly, line 1311: This can be the same thing as L1303.

Kimchelly (approved):
LGTM mod unimportant consistency suggestion.

queue/driver_mongo.go
ybrill: done

PR #317, opened 2023-05-22
EVG-19188 span for jobs

ybrill:
[EVG-19188](https://jira.mongodb.org/browse/EVG-19188)

Create a span for each job execution. If the application has not set a tracer provider in the global state with SetTracerProvider (as [evergreen does](https://github.com/evergreen-ci/evergreen/blob/bffc7e13fbe8bcc053a39f4611d735b464c09d90/environment.go#L885)) the traces will noop. Since the span is recorded in the context and passed to the job any instrumentation in the job will get hooked in under the job's context (including any db calls it makes using the context).

Please flag if there are any other attributes that might be useful. Most of the attributes in HC so far use a sort of dot notation, so it made sense to stick with that here as well.


pool/helpers.go
```diff
@@ -86,11 +103,32 @@ func executeJob(ctx context.Context, id string, j amboy.Job, q amboy.Queue) {
...
 	if err := j.Error(); err != nil {
 		grip.Error(message.WrapError(err, msg))
+		span.SetStatus(codes.Error, "job encountered error")
+		span.RecordError(err, trace.WithStackTrace(true))
 	} else {
 		grip.Info(msg)
 	}
 }
 
+func jobSpan(ctx context.Context, j amboy.Job) (context.Context, trace.Span) {
```
Kimchelly, line 113: Can we mention in a doc comment that `jobSpan` is a no-op for Amboy queues running in applications that don't have Honeycomb hooked up already?

pool/helpers.go
```diff
@@ -86,11 +103,32 @@ func executeJob(ctx context.Context, id string, j amboy.Job, q amboy.Queue) {
...
+	ti := j.TimeInfo()
+	attributes := []attribute.KeyValue{
+		attribute.String(jobIDAttribute, j.ID()),
+		attribute.String(jobNameAttribute, j.Type().Name),
+		attribute.Int64(jobDispatchMSAttribute, ti.Start.Sub(ti.Created).Milliseconds()),
```
Kimchelly, line 118: nit: this can just be `j.TimeInfo().Start` (etc...) since the time info is only used on this one line and won't change once the job is already about to run.

pool/helpers.go
```diff
@@ -86,11 +103,32 @@ func executeJob(ctx context.Context, id string, j amboy.Job, q amboy.Queue) {
...
+		span.RecordError(err, trace.WithStackTrace(true))
 	} else {
 		grip.Info(msg)
 	}
 }
 
+func jobSpan(ctx context.Context, j amboy.Job) (context.Context, trace.Span) {
+	ti := j.TimeInfo()
+	attributes := []attribute.KeyValue{
+		attribute.String(jobIDAttribute, j.ID()),
```
Kimchelly, line 116: I would include the queue ID in the span since it encodes additional information like app server name and queue name.

Kimchelly (commented):
You have to bump the Go version to a later one for `golang.org/x/sys`.

pool/helpers.go
ybrill: done

pool/helpers.go
ybrill: done

ybrill:
I don't know the history behind the govet `check-shadowing` setting. It looked like it's false by default so it seemed safe to remove. 
Is that correct?

Kimchelly (approved):
It is safe to remove `check-shadowing` if it's been made false by default, it used to be true. (Although now that it's removed from just this one repo, this exacerbates the issue where the common files like the golangci-lint YAML are out of sync across the PLT repos.)

ybrill:
Hmm. I can try to remove it as I/we update them for their arch use and go version.

ybrill:
I reverted all the go1.20 changes in favor of merging in the changes from https://github.com/mongodb/amboy/pull/318

PR #318, opened 2023-05-23
EVG-19924: Upgrade go version, remove arch linux, upgrade distros.

johndaniels:
This also excises part of a test that was unreliable on macOS in CI in queue/dispatcher_test.go . It doesn't seem to be the primary purpose of the test so this seems like an ok change to make.


evergreen.yaml
```diff
@@ -193,55 +201,48 @@ tasks:
...
-      RACE_DETECTOR: true
-      MONGODB_URL: https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-4.0.3.tgz
-    run_on:
-      - archlinux-new-large
-    tasks:
-      - ".test"
-
   - name: lint
-    display_name: Lint (Arch Linux)
+    display_name: Lint (Ubuntu 22.04)
```
Kimchelly, line 205: We can just call this `Lint` since the Linux distro isn't that important.

waitForMongo.js
```diff
@@ -0,0 +1,31 @@
+//
```
Kimchelly, line 1: I would put this in a `scripts` subdirectory like Evergreen does to avoid polluting the top-level namespace too much.

ybrill:
I pushed [78e9246](https://github.com/mongodb/amboy/pull/318/commits/78e924655758e93d52dcf86edc18bfaff08de21f) reflecting our slack convo.
The gist of the conversation there is that the test is being flakey because sometimes the lockPing's update happens before `oldStatus` is captured and the sleep doesn't make us pass another lock ping.

The idea is to 1) increase the lockTimeout so the ping takes longer and we get to check it first and 2) move the second call to Dispatch after the `oldStatus` so it gets called right away, as it does in an earlier test.

waitForMongo.js
johndaniels: Fixed

evergreen.yaml
johndaniels: Fine by me.

evergreen.yaml
```diff
@@ -193,55 +201,48 @@ tasks:
...
 
   - name: macos
-    display_name: macOS 10.14
+    display_name: macOS 11.00
     expansions:
-      GOROOT: /opt/golang/go1.16
-      MONGODB_URL: https://fastdl.mongodb.org/osx/mongodb-osx-ssl-x86_64-4.0.3.tgz
+      GOROOT: /opt/golang/go1.20
+      MONGODB_URL: https://fastdl.mongodb.org/osx/mongodb-macos-arm64-6.0.6.tgz
+      MONGOSH_DOWNLOAD: curl https://downloads.mongodb.com/compass/mongosh-1.9.0-darwin-arm64.zip -o mongosh.zip && unzip mongosh.zip
```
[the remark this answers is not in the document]
ybrill: I verified that [the build variant's expansions take precedence](https://github.com/evergreen-ci/evergreen/blob/7c0f8f197b6bc71afd7a5bd2665d8fb11b69a4bc/agent/agent.go#L461-L468).

Kimchelly (approved):

evergreen.yaml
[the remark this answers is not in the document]
johndaniels: Except we want to use DECOMPRESS for the mongodb tgz file.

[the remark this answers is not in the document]
ybrill: Oh right! I forgot about that.

PR #330, opened 2023-08-11
EVG-20527 PutMany jobs

ybrill:
[EVG-20527](https://jira.mongodb.org/browse/EVG-20527)

We observed that inserting many jobs (such as what the populators in evergreen do) takes a while because of all the round trips to the db. PutMany will allow adding many jobs to a queue at once.

It was difficult to decide what to return if there is a mixture of write errors, some duplicate key errors and some not. It seemed like the right thing to do (and the expedient thing for `EnqueueManyUniqueJobs`) was to only return a duplicate job error if the errors were _exclusively_ duplicate key errors. Reading [the driver code](https://github.com/mongodb/mongo-go-driver/blob/7355997933f8522a8a17379918b5cb91ff3674cc/mongo/collection.go#L321-L331) made me relatively confident that at least the driver assumes only one error is ever returned for each inserted document, so it seemed okay to assume duplicate insert will only return a duplicate key error and no others. Worst case we'll start noticing more non-duplicate job errors and we'll know this was a mistake.

I wondered if we wanted to replace the error code check with [the one from the driver](https://pkg.go.dev/go.mongodb.org/mongo-driver/mongo#IsDuplicateKeyError), but left it for now. It wasn't clear to me why the error would be reported in any of the three ways. I asked in slack where the error would be expected, but no one responded 🤷 .

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -60,9 +60,9 @@ func EnqueueUniqueJob(ctx context.Context, queue Queue, job Job) error {
 	return errors.WithStack(err)
 }
 
-// EnqueueUniqueJobs is a generic wrapper for adding jobs to queues (using the
-// PutMany() method), but that ignores duplicate job errors.
-func EnqueueUniqueJobs(ctx context.Context, queue Queue, jobs []Job) error {
+// EnqueueManyUniqueJobs is a generic wrapper for adding jobs to a queue (using the
+// PutMany() method) ignoring duplicate job errors.
+func EnqueueManyUniqueJobs(ctx context.Context, queue Queue, jobs []Job) error {
 	err := queue.PutMany(ctx, jobs)
 
 	if IsDuplicateJobError(err) {
--- a/errors.go
+++ b/errors.go
@@ -88,6 +88,10 @@ func (w WriteErrors) Error() string {
 	return catcher.String()
 }
 
+// Cause returns the primary cause of the WriteError. The primary cause is defined as follows:
+//   - If a non-duplicate-job error is encountered it is returned.
+//   - If every error is a duplicate job error and at least one of them is a duplicate scope error a duplicate scope error is returned.
+//   - If every error is a duplicate job error and none of them is a duplicate scope error a duplicate job error is returned.
 func (w WriteErrors) Cause() error {
 	if len(w.OtherErrors) > 0 {
 		return w
--- a/errors.go
+++ b/errors.go
@@ -73,12 +73,14 @@ func EnqueueManyUniqueJobs(ctx context.Context, queue Queue, jobs []Job) error {
 	return errors.WithStack(err)
 }
 
+// WriteErrors collates errors by error type.
 type WriteErrors struct {
 	DuplicateJobErrors   []error
 	DuplicateScopeErrors []error
 	OtherErrors          []error
 }
 
+// Error returns the string representation of the error.
 func (w WriteErrors) Error() string {
 	catcher := grip.NewBasicCatcher()
 	catcher.Extend(w.DuplicateScopeErrors)
@@ -109,7 +111,11 @@ func (w WriteErrors) Cause() error {
 }
 
 // CollateWriteErrors collates errors into a [WriteErrors].
-func CollateWriteErrors(errs []error) WriteErrors {
+func CollateWriteErrors(errs []error) *WriteErrors {
+	if len(errs) == 0 {
+		return nil
+	}
+
 	var writeErrs WriteErrors
 	for _, err := range errs {
 		if IsDuplicateJobScopeError(err) {
@@ -120,7 +126,7 @@ func CollateWriteErrors(errs []error) WriteErrors {
 			writeErrs.OtherErrors = append(writeErrs.OtherErrors, err)
 		}
 	}
-	return writeErrs
+	return &writeErrs
 }
 
 type duplJobError struct {
--- a/errors.go
+++ b/errors.go
@@ -73,19 +73,17 @@ func EnqueueManyUniqueJobs(ctx context.Context, queue Queue, jobs []Job) error {
 	return errors.WithStack(err)
 }
 
-// WriteErrors collates errors by error type.
-type WriteErrors struct {
-	DuplicateJobErrors   []error
-	DuplicateScopeErrors []error
-	OtherErrors          []error
+type writeErrors struct {
+	duplicateJobErrors   []error
+	duplicateScopeErrors []error
+	otherErrors          []error
 }
 
-// Error returns the string representation of the error.
-func (w WriteErrors) Error() string {
+func (w *writeErrors) Error() string {
 	catcher := grip.NewBasicCatcher()
-	catcher.Extend(w.DuplicateScopeErrors)
-	catcher.Extend(w.DuplicateJobErrors)
-	catcher.Extend(w.OtherErrors)
+	catcher.Extend(w.duplicateScopeErrors)
+	catcher.Extend(w.duplicateJobErrors)
+	catcher.Extend(w.otherErrors)
 
 	return catcher.String()
 }
@@ -94,36 +92,34 @@ func (w WriteErrors) Error() string {
 //   - If a non-duplicate-job error is encountered it is returned.
 //   - If every error is a duplicate job error and at least one of them is a duplicate scope error a duplicate scope error is returned.
 //   - If every error is a duplicate job error and none of them is a duplicate scope error a duplicate job error is returned.
-func (w WriteErrors) Cause() error {
-	if len(w.OtherErrors) > 0 {
+func (w *writeErrors) Cause() error {
+	if len(w.otherErrors) > 0 {
 		return w
 	}
-
-	var duplicateError error
-	if len(w.DuplicateScopeErrors) > 0 {
-		duplicateError = MakeDuplicateJobScopeError(w)
+	if len(w.duplicateScopeErrors) > 0 {
+		return MakeDuplicateJobScopeError(w)
 	}
-	if len(w.DuplicateJobErrors) > 0 {
-		duplicateError = MakeDuplicateJobError(w)
+	if len(w.duplicateJobErrors) > 0 {
+		return MakeDuplicateJobError(w)
 	}
 
-	return duplicateError
+	return nil
 }
 
-// CollateWriteErrors collates errors into a [WriteErrors].
-func CollateWriteErrors(errs []error) *WriteErrors {
+// CollateWriteErrors collates errors into a [writeErrors].
+func CollateWriteErrors(errs []error) *writeErrors {
 	if len(errs) == 0 {
 		return nil
 	}
 
-	var writeErrs WriteErrors
+	var writeErrs writeErrors
 	for _, err := range errs {
 		if IsDuplicateJobScopeError(err) {
-			writeErrs.DuplicateScopeErrors = append(writeErrs.DuplicateScopeErrors, err)
+			writeErrs.duplicateScopeErrors = append(writeErrs.duplicateScopeErrors, err)
 		} else if IsDuplicateJobError(err) {
-			writeErrs.DuplicateJobErrors = append(writeErrs.DuplicateJobErrors, err)
+			writeErrs.duplicateJobErrors = append(writeErrs.duplicateJobErrors, err)
 		} else {
-			writeErrs.OtherErrors = append(writeErrs.OtherErrors, err)
+			writeErrs.otherErrors = append(writeErrs.otherErrors, err)
 		}
 	}
 	return &writeErrs
--- a/errors.go
+++ b/errors.go
@@ -60,6 +60,18 @@ func EnqueueUniqueJob(ctx context.Context, queue Queue, job Job) error {
 	return errors.WithStack(err)
 }
 
+// EnqueueUniqueJobs is a generic wrapper for adding jobs to queues (using the
+// PutMany() method), but that ignores duplicate job errors.
+func EnqueueUniqueJobs(ctx context.Context, queue Queue, jobs []Job) error {
+	err := queue.PutMany(ctx, jobs)
+
+	if IsDuplicateJobError(err) {
+		return nil
+	}
+
+	return errors.WithStack(err)
+}
+
 type duplJobError struct {
 	msg string
 }
--- a/errors.go
+++ b/errors.go
@@ -108,7 +108,7 @@ func (w *writeErrors) Cause() error {
 }
 
 // CollateWriteErrors collates errors into a [writeErrors].
-func CollateWriteErrors(errs []error) *writeErrors {
+func CollateWriteErrors(errs []error) error {
 	if len(errs) == 0 {
 		return nil
 	}
--- a/errors.go
+++ b/errors.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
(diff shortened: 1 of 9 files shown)


pool/mock_queue_test.go
```diff
@@ -55,6 +56,14 @@ func (q *QueueTester) Put(ctx context.Context, j amboy.Job) error {
 	}
 }
 
+func (q *QueueTester) PutMany(ctx context.Context, jobs []amboy.Job) error {
+	catcher := grip.NewBasicCatcher()
+	for _, j := range jobs {
+		catcher.Add(q.Put(ctx, j))
```
Kimchelly, line 62: Can we `catcher.Wrap(q.Put(ctx, j), "job '%s'", j.ID())` so we can track which error produced which error line in the resolved error?

queue/driver.go
```diff
@@ -25,6 +25,8 @@ type remoteQueueDriver interface {
 	GetAllAttempts(ctx context.Context, id string) ([]amboy.Job, error)
 	// Put inserts a new job in the backing storage.
 	Put(context.Context, amboy.Job) error
+	// PutMany inserts new jobs in the backing storage.
+	PutMany(context.Context, []amboy.Job) error
```
Kimchelly, line 29: Comment for how it handles multiple errors across different job insertions.

queue/driver_mongo.go
```diff
@@ -677,24 +677,30 @@ func (d *mongoDriver) GetAllAttempts(ctx context.Context, id string) ([]amboy.Jo
...
-	if err != nil {
-		return errors.Wrap(err, "converting in-memory job to interchange job")
-	}
-
-	ji.Scopes = j.EnqueueScopes()
+	return d.PutMany(ctx, []amboy.Job{j})
+}
 
-	d.addMetadata(ji)
+func (d *mongoDriver) PutMany(ctx context.Context, jobs []amboy.Job) error {
```
Kimchelly, line 683: Can we add a comment on how errors are resolved into a single error? For example, if only a subset of jobs encounter errors (some due to duplicate scopes, some due to duplicate job IDs) and the rest succeed or hit other kinds of errors, which error gets returned?

errors.go
```diff
@@ -60,6 +60,18 @@ func EnqueueUniqueJob(ctx context.Context, queue Queue, job Job) error {
 	return errors.WithStack(err)
 }
 
+// EnqueueManyUniqueJobs is a generic wrapper for adding jobs to a queue (using the
+// PutMany() method) ignoring duplicate job errors.
```
Kimchelly, line 64: Can we describe more explicitly how it behaves if there's a mix of duplicate and non-duplicate jobs amongst all of the enqueued ones? And what happens if some jobs hit a duplicate job error and others hit some other kind of error?

queue/driver_mongo.go
```diff
@@ -677,24 +677,30 @@ func (d *mongoDriver) GetAllAttempts(ctx context.Context, id string) ([]amboy.Jo
...
+		ji.Scopes = j.EnqueueScopes()
+		d.addMetadata(ji)
+		jobInterchanges = append(jobInterchanges, ji)
+	}
 
-	if _, err = d.getCollection().InsertOne(ctx, ji); err != nil {
+	if _, err := d.getCollection().InsertMany(ctx, jobInterchanges, options.InsertMany().SetOrdered(false)); err != nil {
 		if d.isMongoDupScope(err) {
-			return amboy.NewDuplicateJobScopeErrorf("job scopes '%s' conflict", j.Scopes())
+			return amboy.NewDuplicateJobScopeErrorf("job scopes conflict")
```
Kimchelly, line 697: Can we include the subset of IDs of the jobs that produced a scope conflict error? If not, can we list out all the job IDs?

queue/driver_mongo.go
```diff
@@ -677,24 +677,30 @@ func (d *mongoDriver) GetAllAttempts(ctx context.Context, id string) ([]amboy.Jo
...
 		}
 		if isMongoDupKey(err) {
-			return amboy.NewDuplicateJobErrorf("job '%s' already exists", j.ID())
+			return amboy.NewDuplicateJobError("job already exists")
```
Kimchelly, line 700: Same thing as above.

queue/driver_mongo.go
```diff
@@ -677,24 +677,30 @@ func (d *mongoDriver) GetAllAttempts(ctx context.Context, id string) ([]amboy.Jo
...
 		}
 
-		return errors.Wrapf(err, "inserting new job '%s'", j.ID())
+		return errors.Wrap(err, "inserting new jobs")
```
Kimchelly, line 703: Same thing as above.

queue/limited.go
```diff
@@ -112,6 +112,19 @@ func (q *limitedSizeLocal) Put(ctx context.Context, j amboy.Job) error {
...
 
+// PutMany adds jobs to the queue. It returns an error if the queue is not yet
+// opened or if any job of the same name already exists in the queue. If the queue is
+// at capacity, it will block until it can be added or the context is done;
+// waiting for these conditions can cause the other queue operations to also
+// block, so it is not recommended to pass a long-lived context to PutMany.
+func (q *limitedSizeLocal) PutMany(ctx context.Context, jobs []amboy.Job) error {
+	catcher := grip.NewBasicCatcher()
+	for _, j := range jobs {
+		catcher.Add(q.Put(ctx, j))
```
Kimchelly, line 123: Same thing as above - include job ID in the wrapped error.

queue/limited_serializable.go
```diff
@@ -147,6 +147,17 @@ func (q *limitedSizeSerializableLocal) Put(ctx context.Context, j amboy.Job) err
 	return nil
 }
 
+// PutMany adds jobs to the queue, returning an error if the queue is not yet
+// opened or if the job already exists in the queue. If the queue is at
+// capacity, PutMany will fail.
+func (q *limitedSizeSerializableLocal) PutMany(ctx context.Context, jobs []amboy.Job) error {
+	catcher := grip.NewBasicCatcher()
+	for _, j := range jobs {
+		catcher.Add(q.Put(ctx, j))
```
Kimchelly, line 156: Same thing as above - include job ID in the wrapped error.

Kimchelly (changes requested):

Kimchelly (commented):
Gonna also request @julianedwards for a second set of eyes.

pool/mock_queue_test.go
ybrill: done

queue/driver.go
ybrill: done

queue/driver_mongo.go
ybrill: Opted to throw in all the job ids because on the premise that a duplicate key error could show up in a `WriteConcernError` or a `CommandError` we wouldn't know which job caused the error.

queue/driver_mongo.go
ybrill: done

queue/driver_mongo.go
ybrill: done

queue/limited.go
ybrill: done

queue/limited_serializable.go
ybrill: done

errors.go
ybrill: I wasn't sure what more to add here. The comment does say that `EnqueueManyUniqueJobs` ignores duplicate job errors which implies the rest, I think.
Though if you had something specific in mind I can definitely add it.

Kimchelly: Hmm, I think the question was more about how this interacts with each queue's individual implementation choice for handling a mix of duplicate and non-duplicate job errors. I would be okay with a comment as simple as something like a little warning saying since this is enqueueing multiple jobs and each queue handles mixes of errors differently, you may get different behavior depending on which queue you're using.

My main concern is that the remote queue is the special behavior in this case (it has a precedence order when it comes to deciding which error type to return) and the other local queues all behave the same way (i.e. they squash all the duplicate job errors into a generic `catcher.Resolve()` untyped error, so they'll never return a duplicate job error), and having each implementation do its own thing means just have to remember what each implementation does (alternatively, if we could make them behave the same as the remote queue does with regards to mixes of errors, then we wouldn't need all the implementation-specific comments).

queue/driver_mongo.go
```diff
@@ -680,6 +680,12 @@ func (d *mongoDriver) Put(ctx context.Context, j amboy.Job) error {
 	return d.PutMany(ctx, []amboy.Job{j})
 }
 
+// PutMany enqueues multiple jobs on the queue.
+// Returns an error if any of the jobs cannot be inserted. If a non-duplicate-job error
+// is encountered it is returned. If every job is a duplicate job error then if any
```
Kimchelly, line 685: nit: "If every error is a duplicate job error"

queue/driver_mongo.go
```diff
@@ -677,24 +677,40 @@ func (d *mongoDriver) GetAllAttempts(ctx context.Context, id string) ([]amboy.Jo
...
 
-	ji.Scopes = j.EnqueueScopes()
-
-	d.addMetadata(ji)
-
-	if _, err = d.getCollection().InsertOne(ctx, ji); err != nil {
+	if _, err := d.getCollection().InsertMany(ctx, jobInterchanges, options.InsertMany().SetOrdered(false)); err != nil {
+		jobIDs := make([]string, 0, len(jobs))
+		for _, j := range jobs {
+			jobIDs = append(jobIDs, j.ID())
```
julianedwards, line 704: Instead of listing all the job IDs here, would it be more useful to return the [set difference](https://github.com/evergreen-ci/utility/blob/main/slice.go#L88-L108) with the result's [InsertedIDs](https://pkg.go.dev/go.mongodb.org/mongo-driver/mongo#InsertManyResult) slice? Since we are continuing on error, some IDs may have been inserted.

queue/driver_mongo.go
```diff
@@ -677,24 +677,40 @@ func (d *mongoDriver) GetAllAttempts(ctx context.Context, id string) ([]amboy.Jo
...
-	ji, err := registry.MakeJobInterchange(j, d.opts.Format)
-	if err != nil {
-		return errors.Wrap(err, "converting in-memory job to interchange job")
+	return d.PutMany(ctx, []amboy.Job{j})
+}
+
+// PutMany enqueues multiple jobs on the queue.
+// Returns an error if any of the jobs cannot be inserted. If a non-duplicate-job error
+// is encountered it is returned. If every job is a duplicate job error then if any
+// of them is a duplicate scope error a duplicate scope error is returned.
```
julianedwards, line 686: What if we used a catcher to add both duplicate scope and duplicate write errors?

julianedwards (changes requested):

queue/driver_mongo.go
Kimchelly: We can't use a catcher because the catcher destroys all the error types and merges them into one untyped error. This makes it difficult to determine individual jobs errors (for things like `EnqueueUniqueJobs`) if the jobs errored due to all duplicate IDs, all duplicate scopes, a mix of duplicate + other DB errors, etc.

julianedwards: As long as the error priority is clear here, I think that is fine then.

errors.go
ybrill: I think they should all be consistent now.
Idk what you think about where to spell out the behavior in a comment. For now it's on the WriteError's `Cause`

queue/driver_mongo.go
ybrill: done

queue/driver_mongo.go
ybrill: done

ybrill:
idk what y'all think about the consolidation into a `WriteErrors`

Kimchelly (commented):
This is actually the ideal solution I was hoping for, I just wasn't sure if it was possible to extract the mongo driver errors out for each job. Just have some minor questions about the `WriteErrors`.

errors.go
```diff
@@ -72,6 +73,56 @@ func EnqueueManyUniqueJobs(ctx context.Context, queue Queue, jobs []Job) error {
 	return errors.WithStack(err)
 }
 
+type WriteErrors struct {
```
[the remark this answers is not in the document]
ybrill: done

errors.go
```diff
@@ -72,6 +73,56 @@ func EnqueueManyUniqueJobs(ctx context.Context, queue Queue, jobs []Job) error {
...
+	DuplicateJobErrors   []error
+	DuplicateScopeErrors []error
+	OtherErrors          []error
+}
+
+func (w WriteErrors) Error() string {
```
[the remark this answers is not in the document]
ybrill: this is private now.
This is surprising to me: is an error's `Error()` really supposed to be documented?

queue/driver_mongo.go
```diff
@@ -756,117 +753,83 @@ func (d *mongoDriver) getAtomicQuery(jobName string, stat amboy.JobStatusInfo) b
...
-	if wce := dupKeyErrs.writeConcernError; wce != nil {
-		if strings.Contains(wce.Message, index) {
-			return true
-		}
-	}
-
-	for _, werr := range dupKeyErrs.writeErrors {
-		if strings.Contains(werr.Message, index) {
-			return true
+func (d *mongoDriver) toWriteError(err error) error {
```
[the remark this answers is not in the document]
ybrill: done

errors.go
```diff
@@ -72,6 +73,56 @@ func EnqueueManyUniqueJobs(ctx context.Context, queue Queue, jobs []Job) error {
...
+	}
+	if len(w.DuplicateJobErrors) > 0 {
+		duplicateError = MakeDuplicateJobError(w)
+	}
+
+	return duplicateError
+}
+
+// CollateWriteErrors collates errors into a [WriteErrors].
+func CollateWriteErrors(errs []error) WriteErrors {
```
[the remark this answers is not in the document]
ybrill: done

errors.go
```diff
@@ -92,35 +93,37 @@ func (w WriteErrors) Error() string {
...
 
-	return duplicateError
+	return nil
 }
 
-// CollateWriteErrors collates errors into a [WriteErrors].
-func CollateWriteErrors(errs []error) WriteErrors {
-	var writeErrs WriteErrors
+// CollateWriteErrors collates errors into a [writeErrors].
+func CollateWriteErrors(errs []error) *writeErrors {
```
Kimchelly, line 111: Generally, an exported function returning an unexported type is undesirable because the caller can't use it in any way other than for its exported methods. Can we just make this return an `error`?

Kimchelly (approved):
LGTM mod return type suggestion.

errors.go
ybrill: done.
Though I'm not sure I follow: if it's an `error` it's even more limiting

Kimchelly: Some Go linters apparently flag it as difficult to use/discouraged. There's been [some discussions](https://www.reddit.com/r/golang/comments/xurmnl/return_unexported_type/) about this before, and the general gist seems to be "if an exported method needs to return something a user wants to use, exporting just the type name is a good workaround". The Go standard library does this a bunch when they want to expose a type you can look at but can't modify the internals.

julianedwards (approved):
Nice solution!

PR #337, opened 2023-11-06
DEVPROD-1717 upgrade mongo to 7.0

ybrill:
[DEVPROD-1717](https://jira.mongodb.org/browse/DEVPROD-1717)

Upgrade mongodb to 7.0. Upgrade mongosh while we're at it.
The expansions weren't being used for the lint variant so for that one instead of being updated they got removed.


ablack12 (approved):
LGTM assuming we aren't concerned about how this will impact amboy reads

PR #339, opened 2024-05-02
DEVPROD-7013 Remove all instance of job priority from Amboy

hadjri:
This field can be removed as it doesn't currently work and we haven't seen a great need for it.

______
edited to add: 
added [this](https://github.com/evergreen-ci/evergreen/pull/8244) evergreen change as well.


Kimchelly (commented):
I actually meant removing this from Evergreen, not removing it from Amboy.

malikchaya2:
This PR was closed because we weren't ready for it. 
We are now ready for it so I re-opened it and merged main.

Kimchelly (commented):
Just grepping through the changes, it seems like priority is still used in a number of files:
```txt
stats.go
logger/sender_test.go
registry/mock_job_for_test.go
queue/driver_mongo.go
queue/driver_mongo_test.go
```
There's also still a flag in the Evergreen admin settings for it.

Kimchelly:
Remove the driver's support for job priority is part of the ticket. If the job itself can't set priority, then the driver's features that support the job priority functionality need to also be removed because they're now dead code.

malikchaya2:
> Removing the driver's support for job priority is part of the ticket. If the job itself can't set priority, then the queue/driver's features that support the job priority functionality need to also be removed because they're now dead code.
> 
> It also looks like the queue stats don't need the priority removed since that priority field is the logging priority, not the Amboy job priority.

I put back status and removed the driver's support for job priority

queue/driver_mongo.go
```diff
@@ -111,7 +109,6 @@ func DefaultMongoDBOptions() MongoDBOptions {
 	return MongoDBOptions{
```
Kimchelly, line 109: Can't comment on the line, but remove the mention of priority in the doc comment.

queue/driver_mongo_test.go
```diff
@@ -54,7 +54,6 @@ func TestMongoDBOptions(t *testing.T) {
 		t.Run("FailsWithSampleSizeAndPriority", func(t *testing.T) {
 			opts := defaultMongoDBTestOptions()
 			opts.SampleSize = 100
```
Kimchelly, line 56: This test is no longer testing the mutually exclusive options since priority was removed, so the test case can be removed entirely.

Kimchelly (changes requested):
Can also remove [this](https://github.com/mongodb/amboy/blob/37e89d53ded5d5a58d43b557973f24da8ba768e6/queue/driver_mongo.go#L87) and [this](https://github.com/mongodb/amboy/blob/37e89d53ded5d5a58d43b557973f24da8ba768e6/registry/mock_job_for_test.go#L26) since those are also about job priority.

queue/driver_mongo.go
malikchaya2: fixed. And I realized my changes in mock didn't save, fixed it now.

Kimchelly (approved):

PR #341, opened 2024-05-20
DEVPROD-7468 don't recursively return a causer

ybrill:
[DEVPROD-7468](https://jira.mongodb.org/browse/DEVPROD-7468)

If the a db write returns errors that aren't duplicate job/scope errors [this call to Cause](https://github.com/mongodb/amboy/blob/4bafe808c506bf69edb54aeddbfa65bb37062fb0/errors.go#L167) gets into an infinite recursion because the returned error implements `Cause`...

Return [a vanilla error](https://github.com/mongodb/grip/blob/584e10f183651989291528fc432824587903b6fd/errors_multi.go#L246) instead.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -90,7 +90,7 @@ func (w *writeErrors) Error() string {
 }
 
 // Cause returns the primary cause of the WriteError. The primary cause is defined as follows:
-//   - If a non-duplicate-job error is encountered it is returned.
+//   - If a non-duplicate-job error is encountered a plain error type is returned.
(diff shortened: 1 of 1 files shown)


Kimchelly (approved):
LGTM mod can we clarify the first bullet of the comment to say that if any one of the errors is a miscellaneous error (i.e. not a duplicate job error), it'll return all the errors rather than just the miscellaneous one?

ybrill:
changed the first bullet, but a little differently: it now says that it'll return a plain error, rather than the duplicate{whatever} error types

PR #343, opened 2024-05-30
DEVPROD-7058: truncate excessive errors when saving job

Kimchelly:
[DEVPROD-7058](https://jira.mongodb.org/browse/DEVPROD-7058)

The error from the generate.tasks job is that _the job itself_ was too large to save back to the DB when marking it complete, which prevented it from finishing. The only way I could practically see this happening is that the generate.tasks logic produced a huge number of errors that exceeded the MDB doc size limit. The Splunk logs hint that this may be the case, even though they don't directly say it (Splunk has the same issue - it's also unable to log events if the payload is too large).

I opted to truncate the errors when saving them back to the DB. Saving the entire error string back to the job isn't a requirement for Amboy to work correctly, and just knowing that there are errors and having a long snippet of the error is sufficient.

* Truncate errors when saving job back to the DB if the error is excessively long.
    * Warn when the errors are excessively long and need to be cut down.
* Improve logging a little bit when repeatedly trying to mark a job complete (and fix off by one errors in the retry loop).
* Drive-by fix an trivial issue where the job lock pinger logged duration in nanoseconds rather than seconds.

### DIFF
--- a/queue/remote_base.go
+++ b/queue/remote_base.go
@@ -254,21 +254,23 @@ func (q *remoteBase) Complete(ctx context.Context, j amboy.Job) error {
 			if err != nil {
 				if attempt >= maxAttempts {
 					grip.Warning(message.WrapError(err, message.Fields{
-						"job_id":       id,
-						"driver_type":  q.driverType,
-						"job_type":     j.Type().Name,
-						"driver_id":    q.driver.ID(),
-						"num_attempts": attempt,
-						"message":      fmt.Sprintf("after %d attempts, aborting marking job complete", attempt),
+						"message":       fmt.Sprintf("after %d attempts, aborting marking job complete", attempt),
+						"job_id":        id,
+						"driver_type":   q.driverType,
+						"job_type":      j.Type().Name,
+						"driver_id":     q.driver.ID(),
+						"num_attempts":  attempt,
+						"duration_secs": time.Since(startAt).Seconds(),
 					}))
 				} else if amboy.IsDuplicateJobError(err) || amboy.IsJobNotFoundError(err) {
 					grip.Warning(message.WrapError(err, message.Fields{
-						"job_id":       id,
-						"driver_type":  q.driverType,
-						"job_type":     j.Type().Name,
-						"driver_id":    q.driver.ID(),
-						"num_attempts": attempt,
-						"message":      "attempting to complete job without lock",
+						"message":       "attempting to complete job without lock",
+						"job_id":        id,
+						"driver_type":   q.driverType,
+						"job_type":      j.Type().Name,
+						"driver_id":     q.driver.ID(),
+						"num_attempts":  attempt,
+						"duration_secs": time.Since(startAt).Seconds(),
 					}))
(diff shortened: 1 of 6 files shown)


queue/remote_base.go
```diff
@@ -251,40 +252,33 @@ func (q *remoteBase) Complete(ctx context.Context, j amboy.Job) error {
 
 			err = q.driver.Complete(ctx, j)
 			if err != nil {
-				if time.Since(startAt) > time.Minute+q.Info().LockTimeout {
```
ablack12, line 254: Is this version of looping specifically keeping our jobs  < a minute? Do we just not think this is going to be a concern, since even large generators seem to hit 10 retries? Or is this if statement just not doing anything?

registry/interchange.go
```diff
@@ -45,12 +50,22 @@ func MakeJobInterchange(j amboy.Job, f amboy.Format) (*JobInterchange, error) {
 		return nil, err
 	}
 
+	status := j.Status()
+	truncatedErrs, isTruncated := truncateJobErrors(status.Errors)
+	status.Errors = truncatedErrs
+	grip.WarningWhen(isTruncated, message.Fields{
```
ablack12, line 56: do we log status.Errors elsewhere then? I'm skeptical that we should be logging truncated_errs twice. Does this error seem helpful in general? Also should we log this at a debug level if we want to keep it?

ablack12 (changes requested):

registry/interchange.go
Kimchelly: We need this as a monitoring mechanism because if we don't log the truncated errors here, we'll never be able to tell that there was a problem. Because the error message is so long, the generic job log [here](https://github.com/mongodb/amboy/blob/f84bf0b328aee41ff97a31d9a7380a2b8f079a51/pool/helpers.go#L92-L103) will never log to Splunk (for the reason stated in the description), so we'll have no way of telling that the job really completed, nor what errors there were.

queue/remote_base.go
Kimchelly: It's not doing anything, it's just old logging that doesn't really serve a purpose.

registry/interchange.go
ablack12: Should the truncating happen at the time of executing the job instead then? That way the logs will be consistent regardless of the size of error, and then when we get to the job interchanging the status will already be a reasonable size (I think)

Kimchelly: I mean, I guess we could truncate while doing `AddErrors`, but that would also result in weirdness because we wouldn't have a way of informing the user (unless we logged in `AddErrors`, which seems even noisier than just logging when the job is saved back to the DB).

In any case, this is an edge case that only gets hit specifically by generate.tasks in a case where the YAML has so many errors that it emits 16 MB worth of them. It's rare enough that we've only seen this bug occur once in 5 years. I don't think that it's worth putting a lot of effort into handling an edge case for a specific job.

ablack12 (approved):

PR #348, opened 2024-09-04
Revert "feat: upgrade go drivers to v2 beta"

ZackarySantana:
This moves this repo off of the RC for the go beta, once it is released (then we should upgrade)


Kimchelly (commented):
Is this also supposed to revert DEVPROD-7013?

ZackarySantana:
no, that's why we have reviews 🥇 thank you

Kimchelly (approved):

